PR #7, opened 2022-04-20
fix clone url in readme

hdevalence:


maxammann:
I just resolved a conflict, thanks for the fix!

PR #13, opened 2022-04-20
Add gradle project for Android

maxammann:


nyurik:
same as for apple - would it make sense to have `/android` instead for simplicity? At worst we could have sub-dirs with `/android/maplibre` and `/android/something-else`

PR #15, opened 2022-04-20
Refactor XCode project

maxammann:
Refactored xcode project into a framework.

* This framework can be bundled into an xcframework
* The xcframework can be added to a swift package

Deployment happens through #21


nyurik:
do you really want to have `maplibre-apple` inside `maplibre`?  Why not just keep it as `/apple`?  At worst maybe `/apple/maplibre` - this way when typing, its enough to say `cd a<TAB><TAB>` and it will autocomplete both

PR #35, opened 2022-04-24
Experiment with GPU-based font rendering

FabianWildgrube:
Adds an experiments folder with a standalone native app that tests rendering of text on the GPU.

See detailed [explanation and a first benchmark here](https://github.com/maplibre/maplibre-rs/tree/fontrendering/experiments/wgpu_for_fontrendering).

This is not finished, rather pushed to the repo for visibility as it was developed in a private fork.

Open Issues include:
* overlapping text "cancels" each other out, due to trick in fragment shader
* Glyphs are "tesselated" and drawn naively (i.e. each gets their own buffer, each gets a single draw call)
   - Only tesselate each glyph once
   - Draw instanced

---
This experiment could also be refactored to abstract the boilerplate code for the standalone winit/wgpu application away, so that that can be reused for other experiments (if we don't want to always work with the full maplibre-rs project as the basis)


FabianWildgrube:
Close in order to re-open from forked repo to follow project policy

PR #38, opened 2022-04-25
optimize zoom bounds

nyurik:
Minor optimization - no point to compute bounds on every calculation - enough to keep them around with const pre-generation.


maxammann:
Looks good!

As I'm not an expert in this GIS world I have a question: Is this constant expected to be correct for all tile-based renderers?
I was already wondering whether my calculations follow a globally accepted standard for vector tiles, or whether there are different tile-based map renderers which use different bounds.

maxammann:
Looks good!

As I'm not an expert in this GIS world I have a question: Is this constant expected to be correct for all tile-based renderers?
I was already wondering whether my calculations follow a globally accepted standard for vector tiles, or whether there are different tile-based map renderers which use different bounds.

maxammann:
Will merge after main is fixed

maplibre/src/coords.rs
```diff
@@ -139,7 +153,9 @@ impl TileCoords {
     /// The [`TileCoords`] `T(x=5,y=5,z=0)` exceeds its bounds because there is no tile
     /// `x=5,y=5` at zoom level `z=0`.
     pub fn into_world_tile(self, scheme: TileAddressingScheme) -> Option<WorldTileCoords> {
-        let bounds = 2i32.pow(self.z as u32);
+        // FIXME: MAX_ZOOM is 32, which means max bound is 2^32, which wouldn't fit in u32 or i32
+        // Note that unlike WorldTileCoords, values are signed (no idea why)
```
maxammann, line 157: I think we could make WorldTileCoords unsigned.

maxammann (approved):

maplibre/src/coords.rs
nyurik: I got a bit confused by why you have two near-identical implementations here tbh.  Sure, an unsigned would work just fine.

maplibre/src/coords.rs
```diff
@@ -191,7 +207,8 @@ impl WorldTileCoords {
     /// The [`WorldTileCoords`] `WT(x=5,y=5,z=0)` exceeds its bounds because there is no tile
     /// `x=5,y=5` at zoom level `z=0`.
     pub fn into_tile(self, scheme: TileAddressingScheme) -> Option<TileCoords> {
-        let bounds = 2u32.pow(self.z as u32);
+        // FIXME: MAX_ZOOM is 32, which means max bound is 2^32, which wouldn't fit in u32 or i32
```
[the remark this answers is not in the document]
nyurik: well, most mvts are generated up to z14, with some services going up as high as z16 for some areas. z18-z20ish is the zoomed-in building level. I seriously wouldn't imagine anyone needing to go that far beyond that. 

WRT zoom encoding, I have some fun code to encode x/y/z into a single u32 value -- see https://github.com/nyurik/planetiler-rs/blob/main/experiments/src/tile_id.rs -- but that only goes up to z15, so would not be suitable for the non-MVT frontend.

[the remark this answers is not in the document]
maxammann: Thats cool! Let's stay for now with 32, because it is a nice number :)

maplibre/src/coords.rs
maxammann: lets discuss here: https://github.com/maplibre/maplibre-rs/issues/65

PR #40, opened 2022-04-25
Fontrendering

FabianWildgrube:
Adds an experiments folder with a standalone native app that tests rendering of text on the GPU.

See detailed [explanation and a first benchmark here](https://github.com/FabianWildgrube/maplibre-rs/tree/fontrendering/experiments/wgpu_for_fontrendering).

This is not finished, rather pushed to the repo for visibility as it was developed in a private fork.

Open Issues include:

overlapping text "cancels" each other out, due to trick in fragment shader
Glyphs are "tesselated" and drawn naively (i.e. each gets their own buffer, each gets a single draw call)
Only tesselate each glyph once
Draw instanced
This experiment could also be refactored to abstract the boilerplate code for the standalone winit/wgpu application away, so that that can be reused for other experiments (if we don't want to always work with the full maplibre-rs project as the basis)


FabianWildgrube:
Current status: Instances Rendering with support for per-word colors
Screenshot 2022-05-15 at 10 27 29
(This frame took 8ms rendering time on M1 Macbook Pro)

maxammann:
Worth checking out: https://github.com/yeslogic/allsorts

maxammann:
@FabianWildgrube Hi! Are you still interested in pushing this forward? I think this approach is still very nice, though there were some limitations I don't remember right now.

FabianWildgrube:
Hi @maxammann - I still think it's an interesting and fun problem to look at. But realistically I'm not going to get to it because of work and other obligations 😞

maxammann:
> Hi @maxammann - I still think it's an interesting and fun problem to look at. But realistically I'm not going to get to it because of work and other obligations 😞

Yeah same here :) no worries!

PR #55, opened 2022-04-28
Remove the default build target in the Cargo configuration

Drabble:
When I run the command `cargo run -p maplibre-demo` on my Mac (version 11.6.4), the target should be *x86_64-apple-darwin*. 

Because of the default config in `.cargo/config.toml` the target that is used instead is *x86_64-unknown-linux-gnu* and I get the error `error: failed to run custom build command for wayland-sys v0.29.4`.

I have looked at a few major Rust repositories and none seem to use a default cargo target.

I think we should remove the default build target so Rustup will use its active toolchain as shown by `rustup show`.

What do you think?

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,3 @@
-[build]
-target = "x86_64-unknown-linux-gnu"
-
 [target.wasm32-unknown-unknown]
 rustflags = [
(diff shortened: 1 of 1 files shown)


maxammann (approved):
Makes perfect sense! Thanks!

maxammann (approved):

PR #80, opened 2022-05-05
Refactor WebWorker pool

maxammann:
The goal of the PR is to simplify the current worker pool which was copy pasted from some other repository.

## ✔️ PR Todo


web/src/platform/pool.rs
```diff
@@ -139,76 +140,15 @@ impl WorkerPool {
...
-    /// Configures an `onmessage` callback for the `worker` specified for the
-    /// web worker to be reclaimed and re-inserted into this pool when a message
-    /// is received.
-    ///
-    /// Currently this `WorkerPool` abstraction is intended to execute one-off
-    /// style work where the work itself doesn't send any notifications and
-    /// whatn it's done the worker is ready to execute more work. This method is
-    /// used for all spawned workers to ensure that when the work is finished
-    /// the worker is reclaimed back into this pool.
-    fn reclaim_on_message(&self, worker: Worker) {
```
maxammann, line 152: I removed this because we don't need to be notified when work is done.

Maybe we should add this back though. That way we could schedule work to an idle worker.

PR #84, opened 2022-05-06
Write the Microsoft Windows build documentation

Drabble:
Maplibre-rs doesn't have a documentation for Microsoft Windows yet. There are some requirements that need to be installed such as Visual Studio c++ build tools. It is important to test and document it.

We'll probably want to make a Github Action for Windows releases in the future.

## ✔️ PR Todo

- [x] Write the documentation.
- [ ] Visual studio build tools for c++ probably already installs CMake so we can remove this step.
- [ ] Figure out what is the issue with sqlite3.lib when building the app.
- [ ] Test out on a fresh Windows 10 install (maybe Windows 11?).
- [ ] Check for errors and inconsistencies in the documentation.

### DIFF
--- a/docs/src/development-guide/building.md
+++ b/docs/src/development-guide/building.md
@@ -83,6 +83,7 @@ Finally you can run the command:
 
 ```bash
 cargo build -p maplibre-demo --target x86_64-pc-windows-msvc
+```
 
 > The target may vary depending on the CPU architecture you are targetting for your build. On ARM64 you would
 > use the target aarch64-pc-windows-msvc instead. If you are building for your current operating system, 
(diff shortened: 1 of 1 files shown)


maxammann:
Did you already experiment with WSL 1/WSL 2. I think this could simplify the compilation significantly. I would expect it to "just work".

Drabble:
> Did you already experiment with WSL 1/WSL 2. I think this could simplify the compilation significantly. I would expect it to "just work".

That's a good idea, I will probably try it this week-end and I'll let you know!

Drabble:
The documentation in #102 replaces this PR.

PR #85, opened 2022-05-06
Contributors README.md

### LINKED ISSUE #59: Add a section about contributors to README.md
We should mention all the awesome people who make maplibre-rs possible!

## 🤔 Expected Behavior

We should link to twitter/matrix/github profiles. Ideally we would mention roles attached to people who regularly contribute.

There should be an somewhat automatic way to update the contributors.

## 💁 Possible Solution
We could use something like that: https://github.com/all-contributors/all-contributors/blob/master/README.md

neimsaci:
Created initial Contributor list with pictures and links to respective Github accounts

fixes #59


maxammann:
Looks very good! did you somehow generate this automatically? Also should we add it directly to the README?

Maybe we could also add an optinal "role" to each contributor. For example for @FabianWildgrube that would be "Text Rendering Guru". For me its probably "Project Representative / Cross-platform Guru". What do you think?

maxammann:
I think we need some way to update this semi-automatically. Maybe there is a bot for that.

neimsaci:
Bot is great but I have no idea yet how to accomplish that.

FabianWildgrube:
I agree, this should be automated. As the project grows, maintaining this manually will become a real pain.
It might be possible to solve this with a Github Action (e.g., https://github.com/marketplace/actions/auto-add-contributors). I'm not an expert on those (and possible costs incurred when an action is run on every push?) but that might be the quickest way to get this up and running.

maxammann:
I just noticed that Github is already presentation contributors quite nicely:
image

So I think this section of the README should add some more value. So maybe instead of having the goal of listing all contributors, we could list the "core maintainers" or "experts on various topics" and add links to social media (twitter, matrix etc.).

That way new users would know who to contact with questions.

neimsaci:
That's a great idea! However question where in repo e.g. folder Contributors README.md should be? In front page it will duplicate some information that is in the default contributor list on the right.

maxammann:
Closing this, as the issue was not clearly stated in the beginning, sorry!
I put that issue WIP, lets discuss there how we want to implement this.

PR #87, opened 2022-05-08
Improve the Cargo documentation

Drabble:
I have been reading the code for my own comprehension. I took the opportunity to try and improve the overall Cargo documentation of the project. My understanding is still not complete and there may be issues in my comments. I am more than open to criticism and I still need to go through my comments one more time.

I have also added a few FIXME comments where I found potential issues, these I would like to discuss before merging them in the main branch.

There are multiple points to be defined before we can merge:

- Does it make sense to add simple comments like `Shared thread state` on the `[crate::io:shared_thread_state]` module so it looks better in the Cargo documentation? I think it can make sense to have a placeholder in some cases that we can improve when we have more time and knowledge.
- Which comment convention should we use? I have found this one: [Standard library convention](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text).
- Is it useful to add a period at the end of each phrase in comments?

## 🚨 Test instructions

Run the following command to open the documentation:

```bash
cargo doc -p maplibre --no-deps --open
```

## ✔️ PR Todo

- [x] Validate the comments
- [x] Review the FIXME comments


maplibre/src/io/mod.rs
```diff
@@ -1,4 +1,4 @@
-//! Handles IO related processing as well as multithreading.
+//! Handles IO related processing with async Rust.
```
Drabble, line 1:
```suggestion
//! Handles IO related processing as well as multithreading.
```

maplibre/src/lib.rs
```diff
@@ -42,19 +47,34 @@ where
     SM: ScheduleMethod,
     HC: HTTPClient,
 {
+    /// Starts the [`crate::map_state::MapState`] Runnable with the configured event loop.
     pub fn run(self) {
         self.run_with_optionally_max_frames(None);
     }
 
+    /// Starts the MapState Runnable with the configured event loop.
```
Drabble, line 55:
```suggestion
    /// Starts the [`crate::map_state::MapState`] Runnable with the configured event loop.
```

maplibre/src/lib.rs
```diff
@@ -42,19 +47,34 @@ where
...
+    ///
+    /// * `max_frames` - Optional maximum number of frames per second.
     pub fn run_with_optionally_max_frames(self, max_frames: Option<u64>) {
         self.map_state.run(self.event_loop, max_frames);
     }
 }
 
+/// Stores the map configuration before the map's state has been fully initialized.
+///
+/// FIXME: We could maybe remove this class, and store the render_state in an Optional in MapState.
```
Drabble, line 76:
```suggestion
/// FIXME: We could maybe remove this class, and store the render_state in an Optional in [`crate::map_state::MapState`].
```

maplibre/src/platform/mod.rs
```diff
@@ -29,16 +29,21 @@ pub const COLOR_TEXTURE_FORMAT: wgpu::TextureFormat = wgpu::TextureFormat::Bgra8
...
 pub mod schedule_method {
     #[cfg(not(target_arch = "wasm32"))]
     pub use super::noweb::schedule_method::*;
 }
 
-// FIXME: This limit is enforced by WebGL. Actually this makes sense!
-// FIXME: This can also be achieved by _pad attributes in shader_ffi.rs
+/// Minimum WebGPU buffer size
+///
+///
```
Drabble, line 46:
```suggestion
///
```

maplibre/src/coords.rs
```diff
@@ -31,6 +31,10 @@ const fn create_zoom_bounds<const DIM: usize>() -> [u32; DIM] {
     result
 }
 
+/// Represents the position of a node within a quad tree. It stores the number of one of the four
+/// squares for each subdivision of the quad tree from zoom level 0.
+///
+/// TODO: Would it be useful to optimize the quadkey and store the keys on 3 bits instead of 8?
```
maxammann, line 37: Yes that definitely would be an optimization. https://github.com/maplibre/maplibre-rs/issues/90

But wouldn't it be 2 bit?

maplibre/src/coords.rs
```diff
@@ -55,6 +59,7 @@ impl fmt::Debug for Quadkey {
     }
 }
 
+/// Zoom is an exponential scale that defines the zoom level for camera and tiles.
```
maxammann, line 62: There is a distiction between Zoom and a zoom level. Zoom level is an u8.

maxammann: We probably should add a struct for ZoomLevel.

maplibre/src/error.rs
```diff
@@ -3,6 +3,7 @@
 use lyon::tessellation::TessellationError;
 use std::sync::mpsc::SendError;
 
+/// Multiple types of map rendering errors
```
maxammann, line 6:
```suggestion
/// Enumeration of errors which can happen during the operation of the library.
```

It's not only rendering errors.

maplibre/src/coords.rs
```diff
@@ -439,6 +444,7 @@ impl From<Point3<f64>> for WorldCoords {
     }
 }
 
+/// Defines a bounding box on a tiled map with a zoom level and a padding.
```
maxammann, line 447: If we add the type `ZoomLevel`:
```suggestion
/// Defines a bounding box on a tiled map with a [`ZoomLevel`] and a padding.
```

maxammann: This should add an in-doc link.

maplibre/src/input/pan_handler.rs
```diff
@@ -99,6 +99,7 @@ impl PanHandler {
 
         if !self.is_panning {
             // starting to pan
+            // FIXME: This `if` statement is not used
```
maxammann, line 102: If statement can be removed.

maplibre/src/input/tilt_handler.rs
```diff
@@ -44,6 +44,10 @@ impl TiltHandler {
         } else {
             Deg::zero()
         };
+        // FIXME: If the goal is to maintain the key pressed to tilt, then process_key shouldn't
+        // increase the delta_pitch but set it to true/false. Increasing the delta will cause
+        // multiple clicks to cause a bigger tilt if more than one event can occur during a
+        // frame.
```
maxammann, line 50: True, I think we should address this in a refactor of the input module.

maxammann: https://github.com/maplibre/maplibre-rs/issues/91

maxammann: Nice catch!

maplibre/src/io/geometry_index.rs
```diff
@@ -12,6 +14,7 @@ use rstar::{Envelope, PointDistance, RTree, RTreeObject, AABB};
 use crate::coords::{InnerCoords, Quadkey, WorldCoords, WorldTileCoords, Zoom, EXTENT, TILE_SIZE};
 use crate::util::math::bounds_from_points;
 
+/// A quad tree storing the current tiles being rendered.
```
maxammann, line 17:
```suggestion
/// A quad tree storing the currently loaded tiles.
```

maplibre/src/io/geometry_index.rs
```diff
@@ -55,6 +58,11 @@ impl GeometryIndex {
     }
 }
 
+/// Index of tiles which can be of two types: spatial or linear.
+/// Spatial tiles are stored in a multi-dimentional tree which represents their position in the tile.
+/// Linear tiles are simply stored in a vector.
+///
+/// A spatial tile index can improve query performance on tiles.
```
maxammann, line 65:
```suggestion
/// A spatial tile index can theoretically improve query performance on tiles. Practically it could be slower though. The `Spatial` index is experimental and currently unused.
```

Maybe we remove the spatial index per tile in the future. This was an experiment.

maplibre/src/io/static_tile_fetcher.rs
```diff
@@ -13,6 +15,7 @@ static TILES: Dir = include_dir!("$OUT_DIR/extracted-tiles");
 #[cfg(not(static_tiles))]
 static TILES: Dir = Dir::new("/path", &[]);
 
+/// Load PBF files from the local direction `extracted-tiles`.
```
maxammann, line 18:
```suggestion
/// Load PBF files which were statically embedded in the `build.rs`
```

maplibre/src/lib.rs
```diff
@@ -1,3 +1,7 @@
+//! # Maplibre-rs
+//!
+//! A multi-platform library for rendering vector tile maps with WebGPU.
```
maxammann, line 3: We need to extend this in the future. We maybe can put a description in the book and reference it from here via:

```
#![doc = include_str!("description.md")]
```

What do you think?

maplibre/src/lib.rs
```diff
@@ -108,10 +134,24 @@ where
     SM: ScheduleMethod,
     HC: HTTPClient,
 {
+    /// Runs the map application without a maximum number of frames per second defined.
+    ///
+    /// # Panics
+    ///
+    /// * Panics if the operating system is unable to starts the worker threads.
```
maxammann, line 141:
```suggestion
    /// * Panics if the platform is unable to start the worker threads.
```

maplibre/src/lib.rs
```diff
@@ -108,10 +134,24 @@ where
...
 
+    /// Builds and runs the Map with the given maximum number of frames per second.
+    ///
+    /// # Arguments
+    ///
+    /// * `max_frames` - The maxiumum number of frame per seconds.
+    ///
+    /// # Panics
+    ///
+    /// * Panics if the operating system is unable to starts the worker threads.
```
maxammann, line 154:
```suggestion
    /// * Panics if the platform is unable to start the worker threads.
```

maplibre/src/lib.rs
maxammann: Is that really the case? All unwraps in the code currently should basically go away soon. So I think we can avoid document panics.

maplibre/src/lib.rs
```diff
@@ -73,6 +93,12 @@ where
     SM: ScheduleMethod,
     HC: HTTPClient,
 {
+    /// Initializes the whole rendering pipeline for the given configuration.
+    /// Returns the initialized map, ready to be run.
+    ///
+    /// # Panics
+    ///
+    /// * Panics if Winit is unable to retrieve the target window size.
```
maxammann, line 101: Which panic do you refer here to?

maplibre/src/lib.rs
```diff
@@ -183,6 +224,12 @@ where
         self
     }
 
+    /// Builds the UninitializedMap with the given configuration.
+    ///
+    /// # Panics
+    ///
+    /// * Panics if no schedule method has been configured.
+    /// * Panics if no http client has been configured.
```
maxammann, line 232: I think we can leave this undocumented right now. We will remove all unwraps in a future version.

maplibre/src/map_state.rs
```diff
@@ -1,3 +1,5 @@
+//! Stores the state of the map such as zoom level, camera, styles, etc.
```
maxammann, line 1:
```suggestion
//! Stores the state of the map such as [`Zoom`], camera, styles, etc.
```

Maybe we should link here the proper structs.

maplibre/src/map_state.rs
```diff
@@ -47,6 +51,10 @@ impl ViewState {
     }
 }
 
+/// Stores the state of the map, dispatches tile fetching and caching, tessellation and drawing.
+///
+/// FIXME: MapState may not follow the Single-responsibility principle, as it not only stores
+/// the state of the map but also the rendering, caching, etc.
```
maxammann, line 57: I kind of see `MapState` as the central struct which holds all the schedulers, managers, controllers etc. I agree that some methods in here, probably should not be here. They have been moved here by a previous refactoring.

I think we could add controllers which handle the scheduling of fetching tiles for example.

maplibre/src/platform/noweb/http_client.rs
```diff
@@ -6,6 +6,7 @@ use reqwest_middleware::ClientWithMiddleware;
 use reqwest_middleware_cache::managers::CACacheManager;
 use reqwest_middleware_cache::{Cache, CacheMode};
 
+/// Reqwest http client
```
maxammann, line 9:
```suggestion
```

I think we should avoid such comments, because they add no value.

maplibre/src/style/layer.rs
```diff
@@ -11,6 +14,7 @@ pub struct BackgroundPaint {
     // TODO a lot
 }
 
+/// Fill styling.
```
maxammann, line 17:
```suggestion
```

This documentation should also be removed I think.

maplibre/src/winit.rs
```diff
@@ -89,6 +89,10 @@ where
                     let dt = now - last_render_time;
                     last_render_time = now;
 
+                    // FIXME: Inputs are updated in RedrawRequested which
+                    // 1. Adds a strain on the main thread, maybe we need another thread for that.
+                    // 2. Delta time is correlated to the frame rate, and a low frame rate will cause
+                    //    a lag with inputs. e.g. a button is held for a certain amount of time.
```
maxammann, line 95: 1. Using a different thread is not possible for web, because the DOM/Document can only be accessed from the main thread. I'm fairly sure that this is also the case for other platforms.
2. Can we fix that?

maxammann (approved):

maplibre/src/coords.rs
Drabble: I was wrong with the 3 bits idea. I didn't understand that the first 8 bits of the Quadkey store the zoom level. 

Shouldn't the key be `let mut key = [0u8; MAX_ZOOM+1];` if we store the zoom level in the first byte?

Anyway if we store the zoom level separatly we indeed only need 2 bits for NW, NE, SW and SE.

Should I update #90 with `2 bits * MAX_ZOOM + 8 bits for zoom level`?

maxammann: You are right :)

maxammann: Yes you can change it

nyurik: Just FYI, I was hacking on an optimized tile ID storage a while back - see https://github.com/nyurik/planetiler-rs/blob/main/experiments/src/tile_id.rs -- essentially its a single u32 that can store all z/x/y for z in `0..=15`.  I might create a crate at some point that has all the basic ID features.

maplibre/src/winit.rs
Drabble, line 95:
```suggestion
```

Drabble: You are probably right, and Winit probably already does something special to handle inputs.

Another thing, it's probably nothing but I have to check the Winit documentation. Imagine you call `window_input` twice before the `RedrawRequested` event is called. Maybe that could happen with a slow in framerate. You could miss a mouse click because the key press would have been `true` then `false` in-between two frames. The solution for this could be a buffer for inputs. But that is only useful if `Event::WindowEvent` can happen twice for the same input before a `Event::RedrawRequested`

Drabble (commented):
Add more comments and apply suggestions.

maplibre/src/coords.rs
maxammann: I think for this crate we want +15 zoom, but maybe it would be cool to have an enum between both representations.

maplibre/src/io/shared_thread_state.rs
```diff
@@ -1,3 +1,5 @@
+//! Shared thread state.
```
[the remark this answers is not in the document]
maxammann: Both caches are separate, one stores tesselated data (TIleCache) and one raw geometric data (geometry_index). This probably will be refactored in the future to make more sense.

The index_geometry was used in `QueryHandler`, but its current commented. Not sure why though :D

maplibre/src/lib.rs
maxammann: Sounds good! Probably simpler than including files.

maplibre/src/lib.rs
maxammann: I think after this task all panics should vanish: https://github.com/maplibre/maplibre-rs/issues/92

maplibre/src/winit.rs
maxammann: Sounds correct. Maybe we also should not trust winit's assumptions/documentation.
Can you add this comment to https://github.com/maplibre/maplibre-rs/issues/91 ?

maplibre/src/io/shared_thread_state.rs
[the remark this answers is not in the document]
Drabble: Right, it makes sense to use a `geometry_index` for queries. I am not sure it should be within the `ThreadBus` though.

maplibre/src/winit.rs
Drabble: I have added that comment.

maplibre/src/benchmarking.rs
```diff
@@ -1,7 +1,11 @@
+//! Collection of utilities used to perform certain calculations more conveniently.
+
+/// Re-export of the io module.
 pub mod io {
-    pub use crate::io::*;
+    pub use crate::io;
```
Drabble, line 5:
```suggestion
    pub use crate::io::*;
```

maplibre/src/window.rs
```diff
@@ -47,3 +51,31 @@ impl WindowSize {
...
+pub trait FromWindow {
+    fn from_window(title: &'static str) -> Self;
+}
+
+/// Constructor for a canvas.
+pub trait FromCanvas {
+    fn from_canvas(dom_id: &'static str) -> Self;
+}
+=======
+>>>>>>> main
```
Drabble, line 81:
```suggestion
```

maplibre/src/io/shared_thread_state.rs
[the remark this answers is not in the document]
maxammann: The reason for that was that I wanted to do geometry queries on a separate thread and not block the main thread. Maybe that is not needed though.

maplibre-winit/src/input/tilt_handler.rs
```diff
@@ -42,6 +42,10 @@ impl TiltHandler {
         } else {
             Deg::zero()
         };
+        // FIXME: If the goal is to maintain the key pressed to tilt, then process_key shouldn't
```
maxammann, line 45: Can you add the corresponding issues to these FIXMEs?

After that I think this is ready for merge :)

maxammann (approved):

maplibre/src/lib.rs
```diff
@@ -1,3 +1,21 @@
+//! # Maplibre-rs
+//!
+//! A multi-platform library for rendering vector tile maps with WebGPU.
+//!
+//! Maplibre-rs is a map renderer that can run natively on MacOS, Linux, Windows, Android, iOS and the web.
+//! It takes advantage of Lyon to tessellate vector tiles and WebGPU to display them efficiently.
+//! Maplibre-rs also has an headless mode (*work in progress*) that can generate rasters.
+//!
+//! The official guide book can be found here #![doc = include_str!("description.md")]
```
Drabble, line 9:
```suggestion
//! The official guide book can be found [here](https://maxammann.org/maplibre-rs/docs/)
```

Drabble: It doesn't seem to work with the include_str!. I think a simple link to the docs should work fine.

maplibre-winit/src/input/tilt_handler.rs
```diff
@@ -42,6 +42,10 @@ impl TiltHandler {
...
+        // increase the delta_pitch but set it to true/false. Increasing the delta will cause
+        // multiple clicks to cause a bigger tilt if more than one event can occur during a
+        // frame.
```
Drabble, line 48:
```suggestion
```

maxammann (approved):

maxammann (approved):

PR #93, opened 2022-05-10
Add Render Graph from Bevy

maxammann:
Continued with bevy and not rend3 #94  because:

* rend3 uses some unsafe code which should not be needed
* both have good documentation, but bevys seems a little bit more elaborate
* bevy is already more focused on the web
* bevy's renderer is already in the second iteration. So the design should be fine?

## ✔️ PR Todo

- [x] Included links to related issues/PRs
- [x] Support view/texture resizing
- [x] Multisampling and depth texture
- [x] updating buffers
- [x] Surfaces & presenting
- [x] stages


maxammann:
Kind of still in progress, but ready for a first review. I think we should do this review as a pair-review. So we should schedule a call in order to look over this.

maxammann:
Pair reviewed with @lseelenbinder 

Next up is testing this design by implementing labels and headless rendering #110

PR #97, opened 2022-05-12
Initial start of removing unwrap from a Result type

iakev:
## ✔️ PR Todo

### DIFF
--- a/android/src/lib.rs
+++ b/android/src/lib.rs
@@ -23,15 +23,7 @@ pub fn android_main() {
 
 #[no_mangle]
 pub extern "system" fn Java_org_maplibre_1rs_MapLibreRs_android_1main(env: JNIEnv, class: JClass) {
-    let tag = match CString::new("maplibre");
-    let tag = match tag {
-        Ok(tagg) => tagg,
-        Err(e) => (e),
-    };
-    let message = CString::new("maplibre WOORKING");
-    let message = match message {
-        Ok(mess) => mess,
-        Err(e) => (e),
-    };
+    let tag = CString::new("maplibre").unwrap();
+    let message = CString::new("maplibre WOORKING").unwrap();
     ndk_glue::android_log(Level::Warn, &tag, &message);
 }
--- a/android/src/lib.rs
+++ b/android/src/lib.rs
@@ -23,7 +23,15 @@ pub fn android_main() {
 
 #[no_mangle]
 pub extern "system" fn Java_org_maplibre_1rs_MapLibreRs_android_1main(env: JNIEnv, class: JClass) {
-    let tag = CString::new("maplibre").unwrap();
-    let message = CString::new("maplibre WOORKING").unwrap();
+    let tag = match CString::new("maplibre");
+    let tag = match tag {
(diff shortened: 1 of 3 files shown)


android/src/lib.rs
```diff
@@ -23,7 +23,15 @@ pub fn android_main() {
...
+    let tag = match CString::new("maplibre");
+    let tag = match tag {
+        Ok(tagg) => tagg,
+        Err(e) => (e),
+    };
+    let message = CString::new("maplibre WOORKING");
+    let message = match message {
+        Ok(mess) => mess,
+        Err(e) => (e),
+    };
```
maxammann, line 35: I'm quite sure this won't compile because there is an unneccassary `match` in the first line.
Also the return types of the match arms do not match.

This use if `.unwrap` also happens in a dummy-example used to test JNI for android. This code will eventually be removed soonish.
The general idea was to replace uses of `unwrap` by propagating the error type.

maxammann (changes requested):

iakev:
I understand what you mean now. Maybe in functions that return a Result I
can use the ? operator to propagate the error. I don't wish to push changes
since I have made a lot and am still reviewing them. But for the
above functions is this sufficient?
let tag = CString::new("maplibre");

let tag = match tag {
Ok(_) => tag,
Err(e) => Err(e),
};

let message = CString::new("maplibre WOORKING");

let message = match message {
Ok(_) => message,
Err(e) => Err(e),
};

maxammann:
> I don't wish to push changes since I have made a lot and am still reviewing them. But for the

Just go ahead and push the changes, so we can discuss them early on.

> But for the above functions is this sufficient?

There is no benefit of avoiding `.unwrap()` code there because its just a test function I used to prototype JNI support.

> let tag = match tag { Ok(_) => tag, Err(e) => Err(e), };

That won't work because the types of the match arms are still different.

> Maybe in functions that return a Result I can use the ? operator to propagate the error.

Yes, that's the general idea. We probably should also discuss whether we want to have some helper crate like `thiserror`. Never checked the pros/cons.

iakev:
Hello Max,

I have been trying to work out the unwrap features however, I think my
skill set and understanding of Rust are not sufficient for me to continue
on the project. I am sorry for wasting your time. I think I should polish
my Rust concepts before attempting to help again. I thought since it was a
good first issue it would be easier but I admit I am out of my depth. Again
my sincere apologies.

Best regards,
Kevin Mwongera

maxammann:
> I have been trying to work out the unwrap features however, I think my skill set and understanding of Rust are not sufficient for me to continue on the project. I am sorry for wasting your time. I think I should polish my Rust concepts before attempting to help again. I thought since it was a good first issue it would be easier but I admit I am out of my depth. Again my sincere apologies.

No worries! In fact I have to admit that the issues was also badly written by me. It was very unspecific and too open to serve as a good first issue. Also code has been changing to fast the last weeks that it was difficult.

We'd be happy to receive a future contribution from you!

iakev:
Okay understood, will push them as soon as am done going over them once
more.

PR #100, opened 2022-05-17
Improve apple iOS configuration

Drabble:
When we try to run the `./apple/xcode` project on a MacOS x86_64, the build fails because:

1. The build phases try to run cargo with a target of `x86_64-apple-ios-sim` but there is no `sim` target on x86_64. we need to build for x86_64 instead.
2. The library folder for architecture x86_64 - ios simulator was not set in the library search paths.

Additionally we can lower the iOS target version to version 9.

## 💻 Examples

There have been two errors:

1. `Could not find specification for target "x86_64-apple-ios-sim"`
2. `libtool: can't locate file for: -lmaplibre_apple`

## 🚨 Test instructions

Open and run the iOS project with XCode on a Mac x86_64.

## ✔️ PR Todo

- [x] Build for the cargo target `x86_64-apple-ios` target instead of `x86_64-apple-ios-sim`
- [x] Add the library folder for architecture x86_64 - ios simulator in the library search paths.

### DIFF
--- a/apple/xcode/maplibre-rs.xcodeproj/project.pbxproj
+++ b/apple/xcode/maplibre-rs.xcodeproj/project.pbxproj
@@ -480,7 +480,6 @@
 					"$(PROJECT_DIR)",
 				);
 				"LIBRARY_SEARCH_PATHS[sdk=iphoneos*][arch=arm64]" = "$(PROJECT_DIR)/libs/aarch64-apple-ios/release";
-				"LIBRARY_SEARCH_PATHS[sdk=iphoneos*][arch=x86_64]" = "$(PROJECT_DIR)/libs/x86_64-apple-ios/release";
 				"LIBRARY_SEARCH_PATHS[sdk=iphonesimulator*][arch=x86_64]" = "$(PROJECT_DIR)/libs/x86_64-apple-ios/release";
 				"LIBRARY_SEARCH_PATHS[sdk=iphonesimulator*][arch=arm64]" = "$(PROJECT_DIR)/libs/aarch64-apple-ios-sim/release";
 				"LIBRARY_SEARCH_PATHS[sdk=macosx*][arch=arm64]" = "$(PROJECT_DIR)/libs/aarch64-apple-darwin/release";
--- a/apple/xcode/maplibre-rs.xcodeproj/project.pbxproj
+++ b/apple/xcode/maplibre-rs.xcodeproj/project.pbxproj
@@ -267,7 +267,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/bash;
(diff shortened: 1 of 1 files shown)


apple/xcode/maplibre-rs.xcodeproj/project.pbxproj
```diff
@@ -481,6 +482,7 @@
 				);
 				"LIBRARY_SEARCH_PATHS[sdk=iphoneos*][arch=arm64]" = "$(PROJECT_DIR)/libs/aarch64-apple-ios/release";
 				"LIBRARY_SEARCH_PATHS[sdk=iphoneos*][arch=x86_64]" = "$(PROJECT_DIR)/libs/x86_64-apple-ios/release";
```
maxammann, line 484: Should we remove these search paths? What do you think? I think there is no x86_64 iphone.

apple/xcode/maplibre-rs.xcodeproj/project.pbxproj
```diff
@@ -386,6 +386,7 @@
 				);
 				"LIBRARY_SEARCH_PATHS[sdk=iphoneos*][arch=arm64]" = "$(PROJECT_DIR)/libs/aarch64-apple-ios/debug";
 				"LIBRARY_SEARCH_PATHS[sdk=iphoneos*][arch=x86_64]" = "$(PROJECT_DIR)/libs/x86_64-apple-ios/debug";
```
maxammann, line 388: Same her

maxammann (approved):
This looks very good! Thanks!

So we can also claim now hat we support macOS x86 :tada:

apple/xcode/maplibre-rs.xcodeproj/project.pbxproj
Drabble: You're right, should I do this in the pull request?
We should probably support armv7(s) too but I don't know how it works or if it is compatible with arm64. https://docs.elementscompiler.com/Platforms/Cocoa/CpuArchitectures/#:~:text=iOS%20and%20iPadOS,-On%20iOS%20and&text=arm64%20is%20the%20current%2064,the%20A7%20and%20later%20chips.

maxammann: Yes I think you should remove the iphoneos x86 config in this PR.

Drabble: Yes armv7 armv7s and i386 support was dropped by Apple and Rust: https://blog.rust-lang.org/2020/01/03/reducing-support-for-32-bit-apple-targets.html

apple/xcode/maplibre-rs.xcodeproj/project.pbxproj
Drabble, line 388:
```suggestion
```

apple/xcode/maplibre-rs.xcodeproj/project.pbxproj
Drabble, line 484:
```suggestion
```

maxammann (approved):

PR #101, opened 2022-05-17
Create a Windows executable artifact in GHA

Drabble:
We want to generate a Windows executable (.exe) as an artifact in Github Actions.

## 💻 Examples

\-

## 🚨 Test instructions

Run the github workflow called build_demos.

## ✔️ PR Todo

- [x] Create a Github Action that builds a Windows executable in release mode with cargo.


.github/workflows/build_demos.yml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  workflow_dispatch:
+  push:
+
+jobs:
+  run-check:
+    runs-on: windows-2022
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/demo/windows
```
maxammann, line 12: We have two separate workflows right now.
One runs for PRs and one for the man branch.

I think it makes sense if we build windows in both workflows:
* https://github.com/maplibre/maplibre-rs/blob/7371ae56fd1cf42cd1048cb024995a6228d7c2de/.github/workflows/on_main_push.yml
* https://github.com/maplibre/maplibre-rs/blob/7371ae56fd1cf42cd1048cb024995a6228d7c2de/.github/workflows/on_pull_request.yml

.github/actions/demo/windows/action.yml
```diff
@@ -0,0 +1,29 @@
...
+    - uses: ilammy/msvc-dev-cmd@v1 # Provide access to lib.exe
+    - name: Install SQLite
+      shell: powershell
+      run: choco install sqlite -y --params "/NoTools"
+    - name: Build SQLite lib
+      shell: powershell
+      run: |
+        cd "C:\ProgramData\chocolatey\lib\SQLite\tools"
+        lib /DEF:sqlite3.def /OUT:sqlite3.lib /MACHINE:x64
+        echo "SQLITE3_LIB_DIR=C:\ProgramData\chocolatey\lib\SQLite\tools" >> $env:GITHUB_ENV
```
maxammann, line 21: That looks frightening :D

I think we could easily make sqlite an optional dependency. We only need sqlite in test builds actually. I'll create a PR tomorrow maybe.

maxammann (changes requested):

.github/actions/demo/windows/action.yml
Drabble: I struggled for a long time until I got it working. I thought about removing the dependency on rusqlite but I was not sure how. If you do, it will make the setup on Windows a little bit easier.

.github/workflows/build_demos.yml
Drabble: I think it is a good idea. But artifacts take some storage space and I'm not sure about our limitations. The Windows artifact takes around 4MB I think. The MacOS artifact takes 80MB, probably because I don't build it in release mode yet.

.github/actions/demo/windows/action.yml
maxammann: So its very good that we have this working, as we will need it eventually. https://github.com/maplibre/maplibre-rs/issues/107

.github/workflows/build_demos.yml
maxammann: MacOS is huge because it includes maplibre-rs in quite a lot of different architectures for different platforms. So 80MB is totally fine. Its all release mode.

We don't have to worry about storage space. If it runs out we reduce the retention duraiton.

.github/actions/demo/windows/action.yml
Drabble: Let's resolve this conversation as it works for now, we'll remove the sqlite3 step once #107 is done.

.github/workflows/build_demos.yml
Drabble: I have moved the build action to the two workflows.

Drabble:
The build-demo-windows action should pass as soon as #103 is merged on main.

.idea/misc.xml
```diff
@@ -0,0 +1,11 @@
...
+<project version="4">
+  <component name="ProjectType">
+    <option name="id" value="jpab" />
+  </component>
+  <component name="SwUserDefinedSpecifications">
+    <option name="specTypeByUrl">
+      <map />
+    </option>
+  </component>
+</project>
```
maxammann, line 11: Should not be added

apple/xcode/maplibre-rs.xcodeproj/xcshareddata/xcschemes/maplibre-rs.xcscheme
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
maxammann, line 1: Was this added by accident?

maxammann (approved):

maxammann (approved):

maxammann:
Resolved upstream conflicts and enabled auto merge

PR #103, opened 2022-05-17
Add Windows as a target OS in Cargo dependencies

Drabble:
Windows was missing in the list of target OS in the Cargo dependencies. This caused builds to fail on Windows because dependencies were missing.

## 💻 Examples

\-

## 🚨 Test instructions

Using `cargo run -p maplibre-demo` on a Windows OS now works.

## ✔️ PR Todo

- [ ] Add Windows to the target_os list in the cargo dependencies

### DIFF
--- a/maplibre/Cargo.toml
+++ b/maplibre/Cargo.toml
@@ -16,7 +16,7 @@ no-thread-safe-futures = []
 
 
 
-[target.'cfg(any(target_os = "macos", target_os = "ios", target_os = "linux", target_os = "android"))'.dependencies]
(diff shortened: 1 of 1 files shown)


maxammann (dismissed):

maxammann:
Do you think we need to add windows to other places aswell? In the `platorm` module there are references to linux/macos aswell.

Drabble:
> Do you think we need to add windows to other places aswell? In the `platorm` module there are references to linux/macos aswell.

You're right. I forgot to do it in this pull request. I remember doing it when I was testing things on a Windows computer. I'll fix it tomorrow and test it to make sure it's all good.

maxammann (approved):

PR #104, opened 2022-05-17
Create a MacOS application artifact in GHA

Drabble:
We want to generate a MacOS application (.app) as an artifact in Github Actions.

## 💻 Examples

\-

## 🚨 Test instructions

Run the github workflow called build_demos.

## ✔️ PR Todo

- [x] Create a Github Action that builds a MacOS application xcodebuild.


.github/actions/demo/macos/action.yml
```diff
@@ -0,0 +1,28 @@
...
+  steps:
+    - uses: extractions/setup-just@v1
+      env:
+        GITHUB_TOKEN: ${{ inputs.GITHUB_TOKEN }}
+    - name: Install toolchain
+      shell: bash
+      run: just default-toolchain
+    - uses: Swatinem/rust-cache@v1
+    # TODO: Build in release instad of debug doesn't work
+    # TODO: Sign app
```
maxammann, line 20: I think we should not sign our demos. We should find a workaround for this. We do not want to use non-personal certificates.
Can you create an issue?

.github/workflows/build_demos.yml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  workflow_dispatch:
+  push:
+
+jobs:
+  run-check:
+    runs-on: macos-11
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/demo/macos
```
maxammann, line 12: I think we need to discuss which workflows should exist. We could create two workflows for example. One for the demos and one for the libraries.

maxammann (approved):

.github/actions/demo/macos/action.yml
Drabble: What could be the workaround for this? I can remove the comment but I don't think we can prevent macOS from warning about security risks without signing.

.github/workflows/build_demos.yml
Drabble: Is it ok if I move it to the two current workflows, like for the Windows demo?

- https://github.com/maplibre/maplibre-rs/blob/7371ae56fd1cf42cd1048cb024995a6228d7c2de/.github/workflows/on_main_push.yml
- https://github.com/maplibre/maplibre-rs/blob/7371ae56fd1cf42cd1048cb024995a6228d7c2de/.github/workflows/on_pull_request.yml

.github/actions/demo/macos/action.yml
Drabble:
```suggestion
```

PR #102, opened 2022-05-17
Create the how-to-run.md documentation

Drabble:
There was confusion in the `building.md` file whether the goal was to build an executable binary, run the demos or package the library so it could be imported into other projects.

I have created the `how-to-run.md` file in order to document how to run the demo projects on every single platform. This will be important for developers to test the features they work on. This is also useful for testing purposes and for showcasing the project.

We should probably create another documentat called `how-to-build-as-library.md` that will explain how to build the library in order to include it into other applications. This could include topics such as: how to build a WASM binary that can be used in webapps, how to build a binary + header file that can be used in other languages such as C/C++, or how to package the project to be used in Android and Swift applications... The documentation of this file will probably be linked with github actions that will automate the generations of those libraries/packages.

## 💻 Examples

\-

## 🚨 Test instructions

Execute the steps described in the `how-to-run.md` file to run the demos on various platform.

## ✔️ PR Todo

- [x] Rewrite the documentation in `how-to-run.md`
- [x] Remove the documentation file `building.md`

PR #105, opened 2022-05-17
Android improvements

Drabble:
This PR includes multiples improvements for Android:

1. Add x86 as a target in addition to x86_64 and arm64.
2. Unify minSDK to 21 which is the lowest version that support OpenGL ES 3.1 on Android.
3. Unify targetSDK to 31.
4. Update [https://github.com/mozilla/rust-android-gradle](https://github.com/mozilla/rust-android-gradle) to version 0.9.3 and remove the custom config related to [this issue](https://github.com/mozilla/rust-android-gradle/issues/91) since it has been fixed in the new version. This custom configuration didn't work on non-linux operating systems.
5. Remove limits and leave only `downlevel_defaults` for Android because `downlevel_defaults` has the defaults for OpenGL ES 3.1.

## 💻 Examples

\-

## 🚨 Test instructions

Run the Android demo project with Android Studio. The documentation has been improved in #102.

## ✔️ PR Todo

- [x] Make all the improvements in the list above.


Drabble:
There is an error in the github action for Android. Here is the relevant part of the log:

> Task :lib:cargoBuildX86
error[E0463]: can't find crate for `core`
  |
  = note: the `i686-linux-android` target may not be installed
  = help: consider downloading the target with `rustup target add i686-linux-android`
  = help: consider building the standard library from source with `cargo build -Zbuild-std`

The command `./gradlew assembleDebug` in `just build-android` probably builds for all the targets. The new target x86 doesn't have access to the right toolchain. I will try to fix it tomorrow, and I'll add check for x86 in the github action.

maxammann:
The toolchain needs to be added here: https://github.com/maplibre/maplibre-rs/blob/8b5a4d241fd8e8e5690e0804ec8aa2268d59f069/justfile#L35

android/gradle/.android/advancedFeatures.ini
```diff
@@ -0,0 +1 @@
+GLESDynamicVersion = on
```
maxammann, line 1: How is that file used? I think this is for the emulator, right? how does my local installation find this?

maxammann (approved):
Looks good! thanks!

android/gradle/.android/advancedFeatures.ini
Drabble: This file is needed for the emulator to enable OpenGL ES 3.1 I believe.
I found the solution here: https://stackoverflow.com/a/44552548/2794211

Drabble:
> The toolchain needs to be added here:
> 
> https://github.com/maplibre/maplibre-rs/blob/8b5a4d241fd8e8e5690e0804ec8aa2268d59f069/justfile#L35

I have updated the justfile to install the missing toolchain. Why are we installing it from nightly though?

android/gradle/.android/advancedFeatures.ini
maxammann: But this file should be in `~/.android/`. That means users need to install this manually/make the changes manually. So we need to document this but not add this file.

maxammann:
> I have updated the justfile to install the missing toolchain. Why are we installing it from nightly though?

Because we use a nightly feature which rebuilds the stdlib.

maxammann (changes requested):
.

android/gradle/.android/advancedFeatures.ini
Drabble: Oh that's right, I had read it wrong! My problem solved itself when I added the file, but if the folder it should live in is ~/.android, it is probably something else that solved my problem... I will try again to see if it works without the file and I guess we can remove it if it doesn't help. I'll see if there is another special configuration I made for the emulator in Android Studio and document it. I'll do that next week probably.

Drabble: I have removed the file as it was not used anyway.

maxammann (approved):
Looks good!

PR #114, opened 2022-05-25
Remove tilejson fork

nyurik:
TileJSON code is not being used anywhere, so removing it to keep things simple.  In the future, lets try to use [tilejson crate](https://crates.io/crates/tilejson), which is nearly identical to the code being deleted.


maxammann (approved):
Looks good, if we need it we can add it again later on

Issue #108, opened 2022-05-18
"cargo test" fails with "android-ndk-sys only supports compiling for Android"


dmarcoux:
I'm trying to work on one of the issues labelled with `good first issue` and to ensure that my changes aren't breaking anything, I would like to run the tests, but it's failing.

## 🤔 Expected Behavior

I expect the tests to be executed when I run `cargo test`.

## 😯 Current Behavior

I forked the repository and installed the toolchain with `cargo --version > /dev/null` as seen in the [justfile](https://github.com/maplibre/maplibre-rs/blob/c3edea26e58feabab3df04b6d77ea15cfeecba4d/justfile#L27-L29). I then executed `cargo test` and it failed with the error:

```
error: android-ndk-sys only supports compiling for Android
  --> /home/johndoe/.cargo/registry/src/github.com-1ecc6299db9ec823/ndk-sys-0.2.2/src/lib.rs:22:1
   |
22 | compile_error!("android-ndk-sys only supports compiling for Android");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `ndk-sys` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed
```

## 🌍 Your Environment

* Desktop
    - Operating System: Linux (NixOS)
    - Display Server: Xorg

maxammann:
I think you might uncovered an additional bug. We currently do not run the tests in the CI :see_no_evil: So maybe the tests don't even run :dagger: 

The error above happens because by default `cargo test` runs all tests in all crates in the workspace. But that won't work because at least one crate only works when the target is set to android.
So for now you can just do `cargo test -p maplibre`

dmarcoux:
Oh, I understand now. Thank you for the explanation. Should we update the `justfile` from `cargo test` to `cargo test -p maplibre`?

maxammann:
@dmarcoux that sounds good! hm, but we will also have tests in `maplibre-demo`, "maplibre-android", `apple`, `web` etc.

So we need some way of easily running all the tests.

maxammann:
So I'm not sure how we should handle the ndk dependency. Any crate which includes this can no longer (easily) be used to run tests I think.

maxammann:
[...]

PR #117, opened 2022-05-28
Fix unit test build and run tests on CI (on some platforms)

yzsolt:
Fixes tests so that `cargo test -p maplibre` builds and runs without errors. 
Also enables test running on Windows and Linux CI. Could be enabled for macOS too, but the CI build there's not a simple `cargo build` so I better not touched it.

Some notes:
- There's a rendering test which initializes wgpu. Hopefully it succeeds on CI. If not, I think a separate feature flag for rendering tests should be introduced (and they should be disabled on CI).
- Testing on other platforms (e.g. mobile and web) is far from trivial as they need an actual device/emulator to run on. If there'll be platform-independent code in platform-specific projects, they can be run on CI, but their test builds should be set up separately (with different targets and such).

Partially fixes https://github.com/maplibre/maplibre-rs/issues/108

## ✔️ PR Todo


maplibre/src/render/resource/tracked_render_pass.rs
```diff
@@ -202,8 +202,8 @@ impl<'a> TrackedRenderPass<'a> {
     /// Push a new debug group over the internal stack. Subsequent render commands and debug
     /// markers are grouped into this new group, until [`pop_debug_group`] is called.
     ///
-    /// ```
-    /// # fn example(mut pass: bevy_render::render_phase::TrackedRenderPass<'static>) {
+    /// ```ignore
```
maxammann, line 205: Do you think we can fix these doc-tests? Should I take a look?

maxammann (approved):
Looks good! thanks for fixing the tests :sweat_smile: Didn't pay attention in the last weeks.

* For iOS/macOS: Lets run the test suite for the most prominent target: aarch64-darwin-ios
* For Android: This is tricky because we are cross compiling. We might open a task for this.
* For Web: Tricky aswell, not sure if we can run tests as wasm32. We would need some runtime for that.

maxammann:
This should be the next task we merge as status check names have changed.

Let me know if you want to address any of the comments in the PR or some other. The latter is preferred I think.

maplibre/src/render/resource/tracked_render_pass.rs
yzsolt: These fail because they're using private types. I think that unless you want to make them public and expose them to the end user, the only viable solution would be to refactor them out to a separate, internal library. In the current state of the project however it may not worth doing it, as the architecture will change a lot probably. So in general, not much effort should be put into documenting internal things in-depth (with examples) currently, and ignoring some of these is fine.

maxammann: Ah yes, true, I think in another PR I just made them public. I'll take care!

This documentaiton was 0-effort because it was copied from the great bevy engine :)

PR #130, opened 2022-06-09
Allow panning with touch inputs

Drabble:
It is currently impossible to do anything with touch inputs, on Android at least. The problem is caused by the way we handle `start`, `move` and `end` touch events.

On Android (probably iOS too), the only events we get are:

- Touch start
- Touch move 
- Touch end

For now I have simply fixed the `process_touch_start` method to set the `start_window_position` so that the `update_state` function of `PanHandler` has something to work with. We will have to change the way we process panning and zooming (pinch) to support multiple fingers later.

## 💻 Examples

-

## 🚨 Test instructions

Run on Android, you can now move around with the touch. Zoom is not supported yet with touch inputs.

## ✔️ PR Todo

- [x] Fix panning
- [x] Check that it works on iOS too


maxammann (approved):
Looks good!

maxammann:
The format of one file is not correct

Drabble:
I have tested on iOS but I get an error: 

```
Caused by:
    In a RenderPass
      note: encoder = `<CommandBuffer-(0, 1, Metal)>`
    In a pass parameter
      note: command buffer = `<CommandBuffer-(0, 1, Metal)>`
    attachments have differing sizes: ("color", Extent3d { width: 1170, height: 2532, depth_or_array_layers: 1 }) is followed by ("resolve", Extent3d { width: 1170, height: 2289, depth_or_array_layers: 1 })
```

The error is probably unrelated, I will investigate later.

maxammann (approved):

Issue #116, opened 2022-05-25
Simplify windows build action by not using the `sqlite` feature.


maxammann:
TODO

maxammann:
done in https://github.com/maplibre/maplibre-rs/pull/120

Issue #107, opened 2022-05-18
Hide SQLite3 dependency behind a feature flag


maxammann:
Right now we depend on sqlite3 as a build dependency. This feature is usually only used when embedding tiles statically in the binary for testing.

## 🤔 Expected Behavior

A feature flag should exclude the sqlite3 dependency.

## 😯 Current Behavior

sqlite3 is a hard dependency.

## 🔦 Context

We can simplify parts of the windows CI. For example the demos do not need sqlite3:

https://github.com/maplibre/maplibre-rs/pull/101#discussion_r875234474

PR #131, opened 2022-06-11
Fix feature flag

jackson211:
There is a bug, where I forgot to include `#[cfg(feature = "embed-static-tiles")]` to the method call in the build file, after `sqlite` feature moved away. 
`maplibre-demo ` can not be running since `embed-static-tiles` depends on `sqlite` flag. 

## ✔️ PR Todo

- [x] Included links to related issues/PRs
#107 
#116


maxammann:
Ups, seems like i was too foece-ful with that previous merge :D

maxammann (approved):

jackson211:
It's more like my mistake, should have checked in the previous PR. :)

Issue #132, opened 2022-06-14
Resizing the window causes a panic


Drabble:
When the window is resized the application panics. This is an issue with the way `resize` is called on the surface.

Here is the error output :

```
Caused by:
    In a RenderPass
      note: encoder = `<CommandBuffer-(0, 1, Metal)>`
    In a pass parameter
      note: command buffer = `<CommandBuffer-(0, 1, Metal)>`
    attachments have differing sizes: ("color", Extent3d { width: 1170, height: 2532, depth_or_array_layers: 1 }) is followed by ("resolve", Extent3d { width: 1170, height: 2289, depth_or_array_layers: 1 })
```

## 🤔 Expected Behavior

The window should resize without any error.

## 😯 Current Behavior

The application panics when the window is resized.

## 💁 Possible Solution

Pull request is coming soon.

## 🔦 Context

This issue appears when resizing a window. As on iOS the window is resized on startup, it crashes directly.

## 💻 Code Sample

\-

## 🌍 Your Environment

Any environment except headless.

_Originally posted by @Drabble in https://github.com/maplibre/maplibre-rs/issues/130#issuecomment-1153133761_

PR #133, opened 2022-06-14
Fix window resizing in headed mode

Drabble:
In headed mode, when `resize` is called in `surface.rs`, we update the `surface_config` width and height. Because of that, the`reconfigure` method of the `surface.rs` does not detect any change in the width and height and doesn't call `window.configure`, which is required to apply the `surface_config` to the actual surface.

I propose that we only change `surface.rs` width and height in `resize`, and the actual `surface_config` and `configure` method are updated in the `reconfigure` method called in the render Resource stage.

Related issue #132 

## 💻 Examples

This is a bug fix to allow resizing windows.

## 🚨 Test instructions

Try to resize the window and it should work.

## ✔️ PR Todo

\-

### DIFF
--- a/maplibre/src/render/resource/surface.rs
+++ b/maplibre/src/render/resource/surface.rs
@@ -225,19 +225,14 @@ impl Surface {
 
     pub fn resize(&mut self, width: u32, height: u32) {
         self.size = WindowSize::new(width, height).expect("Invalid size for resizing the surface.");
-        match &mut self.head {
-            Head::Headed(window) => {
-                window.surface_config.height = height;
-                window.surface_config.width = width;
-            }
-            Head::Headless(_) => {}
-        }
     }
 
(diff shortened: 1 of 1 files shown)


maplibre/src/render/resource/surface.rs
```diff
@@ -225,19 +225,14 @@ impl Surface {
...
-            Head::Headless(_) => {}
-        }
     }
 
     pub fn reconfigure(&mut self, device: &wgpu::Device) {
         match &mut self.head {
             Head::Headed(window) => {
                 if window.has_changed(&(self.size.width(), self.size.height())) {
+                    window.surface_config.height = self.size.height();
+                    window.surface_config.width = self.size.width();
```
maxammann, line 235: Maybe we should create a function: `window.resize_and_configure(width, height, device)`

maxammann (approved):

maxammann (approved):

PR #134, opened 2022-06-15
Reduce znear so the map is not cut when we tilt the camera

Drabble:
When we tilt the camera using the R and F keys, the map clips. If we decrease the near clip plane to 1 it no longer does that. 

Fixing this highlights another problem which should be resolved separately in another pr/issue. When tilting the screen we get `thread 'main' panicked at 'Error in Queue::write_buffer: copy of 0..4216 would end up overrunning the bounds of the Destination buffer of size 2176'`. We should: limit the data we write to the buffer; display a lower zoom level for far away tiles when we tilt the screen.

## 💻 Examples

Here is a screen shot with the map clipping due to a big znear value.

Screenshot 2022-06-15 at 12 53 45

## 🚨 Test instructions

Run the maplibre demo and press R or F.

## ✔️ PR Todo

\-

### DIFF
--- a/maplibre/src/context.rs
+++ b/maplibre/src/context.rs
@@ -25,8 +25,13 @@ impl ViewState {
             window_size.width(),
             window_size.height(),
             cgmath::Deg(110.0),
-            1.0,
-            2000.0,
+            // in tile.vertex.wgsl we are setting each layer's final `z` in ndc space to `z_index`.
+            // This means that regardless of the `znear` value all layers will be rendered as part
+            // of the near plane.
+            // These values have been selected experimentally:
+            // https://www.sjbaker.org/steve/omniv/love_your_z_buffer.html
+            1024.0,
+            2048.0,
         );
 
         Self {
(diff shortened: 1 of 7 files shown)


maxammann:
* The out of bounds error is weird and should never happen. I can take a look at it.
* The znear value should not be too small. As far as I know a too small znear to too large zfar value will cause issues with the z buffer: https://en.wikipedia.org/wiki/Z-fighting
  Another turning knob compared to the znear option is the camera height/fovy.

We should maybe check 1-2 resoruces what good values for the znear is.

maplibre/src/context.rs
```diff
@@ -25,7 +25,7 @@ impl ViewState {
             window_size.width(),
             window_size.height(),
             cgmath::Deg(110.0),
-            100.0,
+            1.0,
```
maxammann, line 28: Let's reevaluate what we should choose here. Maybe we should just decrease fovy, which will then increase the camera height.

maxammann:
@maxammann test and merge/approve

maxammann:
I adjusted the znear and zfar. Indeed znear should be as large as possible. I added a reference link.

The actual problem was not the clipping on the GPU, but clipping in the logic of tile loading.

maxammann:
Closing in favor of: https://github.com/maplibre/maplibre-rs/pull/148

PR #135, opened 2022-06-15
Improve Mapbox Style support

Drabble:
Currently we only support a small subset of the [Mapbox Style Specification](https://docs.mapbox.com/mapbox-gl-js/style-spec/root/). In the future, we would like to improve this support and to dynamically load styles from files.

## 💻 Examples

\-

## 🚨 Test instructions

\-

## ✔️ PR Todo

- [x] Add root level elements support such as center (in latitude/longitude), zoom and pitch

### DIFF
--- a/maplibre/src/context.rs
+++ b/maplibre/src/context.rs
@@ -1,4 +1,4 @@
-use crate::coords::{Zoom, ZoomLevel, TILE_SIZE};
+use crate::coords::{Zoom, ZoomLevel, TILE_SIZE, WorldCoords};
 use crate::io::tile_repository::TileRepository;
 use crate::render::camera::{Camera, Perspective, ViewProjection};
 use crate::util::ChangeObserver;
@@ -12,11 +12,14 @@ pub struct ViewState {
 }
 
 impl ViewState {
-    pub fn new(window_size: &WindowSize) -> Self {
+    pub fn new(window_size: &WindowSize, style: &Style) -> Self {
+        let zoom = style.zoom.map_or_else(|| ChangeObserver::default(), |zoom| ChangeObserver::new(Zoom::new(zoom)));
+        let (lat, lon) = style.center.map_or((0.0, 0.0), |center| (center[0], center[1]));
+        let position = WorldCoords::from_lat_lon(lat, lon, zoom.0);
         let camera = Camera::new(
-            (TILE_SIZE / 2.0, TILE_SIZE / 2.0, 150.0),
+            (position.x, position.y, 150.0),
             cgmath::Deg(-90.0),
-            cgmath::Deg(0.0),
+            cgmath::Deg(style.pitch.unwrap_or(0.0)),
             window_size.width(),
             window_size.height(),
         );
@@ -30,7 +33,7 @@ impl ViewState {
         );
 
         Self {
-            zoom: ChangeObserver::default(),
+            zoom,
             camera: ChangeObserver::new(camera),
             perspective,
         }
--- a/maplibre/src/context.rs
+++ b/maplibre/src/context.rs
@@ -1,4 +1,4 @@
-use crate::coords::{WorldCoords, Zoom, ZoomLevel, TILE_SIZE};
+use crate::coords::{LatLon, WorldCoords, Zoom, ZoomLevel, TILE_SIZE};
 use crate::io::tile_repository::TileRepository;
 use crate::render::camera::{Camera, Perspective, ViewProjection};
 use crate::util::ChangeObserver;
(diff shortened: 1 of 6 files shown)


maplibre/src/context.rs
```diff
@@ -12,11 +12,19 @@ pub struct ViewState {
 }
 
 impl ViewState {
-    pub fn new(window_size: &WindowSize) -> Self {
+    pub fn new(window_size: &WindowSize, style: &Style) -> Self {
```
maxammann, line 15: I would favor just passing the `zoom` and `center` manually.

maplibre/src/coords.rs
```diff
@@ -115,7 +116,7 @@ impl Into<u8> for ZoomLevel {
 /// `Zoom` is an exponential scale that defines the zoom of the camera on the map.
 /// We can derive the `ZoomLevel` from `Zoom` by using the `[crate::coords::ZOOM_BOUNDS]`.
 #[derive(Copy, Clone, Debug)]
-pub struct Zoom(f64);
+pub struct Zoom(pub(crate) f64);
```
maxammann, line 119: Should not be public.

maplibre/src/coords.rs
```diff
@@ -460,6 +461,21 @@ fn tiles_with_z(z: u8) -> f64 {
...
+
+        // Convert from degrees to radians
+        let lat_rad = (latitude * PI) / 180.0;
+
+        // get y value
+        let merc_n = f64::ln(f64::tan((PI / 4.0) + (lat_rad / 2.0)));
+        let y = (tile_size / 2.0) - (tile_size * merc_n / (2.0 * PI));
+
+        WorldCoords { x, y }
+    }
```
maxammann, line 477: Should we already make this more abstract? could this differ between different projections?

Also we should add a type for long/lat.

maxammann (commented):
We should decide now or soon how we want to handle the style. I think we should not try to just support the maplibre style, but put more design thought into it now.

maplibre/src/coords.rs
maxammann: OR: create issue about discussing "projections". Is maplibre-rs projection aware?

maxammann (approved):

maplibre/src/coords.rs
maxammann: @Drabble can you create such an issue?

maplibre/src/context.rs
Drabble: I have separated them but I also had to add pitch in the parameters. Should pitch have its own struct too?

maplibre/src/coords.rs
Drabble: I have made this private.

maplibre/src/coords.rs
```diff
@@ -460,6 +461,21 @@ fn tiles_with_z(z: u8) -> f64 {
 }
 
 impl WorldCoords {
+    pub fn from_lat_lon(latitude: f64, longitude: f64, zoom: f64) -> WorldCoords {
```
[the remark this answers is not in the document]
Drabble: I have changed it. I also create a struct called LatLon for now. https://docs.opengeospatial.org/cs/17-002r1/17-002r1.html

maplibre/src/coords.rs
Drabble: #139

nyurik:
I understand the desire to innovate and revisit the boring old specs,and I even (in part) agree, but we should also consider existing users and their ecosystems, and huge investment many companies have already done into designing their own maps.  If maplibre-rs is to become the awesome new tech we all want it to be, there should be an easy path to adaption - and that means minimal migration pain, instead of rewriting all styling from scratch.  I would recommend (eventually) implementing full support of the current styling language, possibly obsoleting a few bad design decisions if we can prove they were bad.  If needed, we could have a new styling language plus some optional compatibility layer - but again, must show we are breaking it for a good reason.

maplibre/src/context.rs
maxammann: Looks fine to me. In some other PR I will change the type of the `pitch` though. It should not be f64 but something that can be turned into a `Rad<f64`.

maxammann:
> I understand the desire to innovate and revisit the boring old specs,and I even (in part) agree, but we should also consider existing users and their ecosystems, and huge investment many companies have already done into designing their own maps. If maplibre-rs is to become the awesome new tech we all want it to be, there should be an easy path to adaption - and that means minimal migration pain, instead of rewriting all styling from scratch. I would recommend (eventually) implementing full support of the current styling language, possibly obsoleting a few bad design decisions if we can prove they were bad. If needed, we could have a new styling language plus some optional compatibility layer - but again, must show we are breaking it for a good reason.

Lets discuss here: https://github.com/maplibre/maplibre-rs/discussions/150

maxammann (approved):

PR #136, opened 2022-06-19
DO NOT MERGE: Refactor `RequestStage` and `PopulateTileStoreStage` stages

Drabble:
Reading through the code, I found that `RequestStage` and `PopulateTileStoreStage` contain a lot of complexity. I have started making a diagram detailing how it works. It felt too complicated, in my opinion, so I tried to refactor it.

Screenshot 2022-06-19 at 11 21 10

I believe this part of the code could be simplified. As a library, maplibre-rs should thrive to be as simple and as easy to use as possible for developers, while still being as customisable as possible.

I believe there are several issues with the design, some of which may have already been discussed. Many of these issues concern the SharedThreadState struct:

```rust
#[derive(Clone)]
pub struct SharedThreadState {
    pub tile_request_state: Arc<Mutex<TileRequestState>>,
    pub message_sender: mpsc::Sender<TessellateMessage>,
    pub geometry_index: Arc<Mutex<GeometryIndex>>,
}
```

1. SharedThreadState doesn't really respect the SRP because it stores the state and it also handles the processing of the tile.

2. SharedThreadState stores the `tile_request_state`, which is a second data structure that stores tiles, in addition to `tile_repository`. I believe the 2 could be merged.
> I have not figured out what GeometryIndex does yet, I believe it is used for querying. Could you explain it to me? It may require some more work on my part to refactor correctly.

3. SharedThreadState contains 2 atomically reference counted  mutexes. This is not very good for performance. We can make it work with the mpsc only.

4. RequestStage and PopulateTileStoreStage depend on eachother, and will likely never be separated. I believe it doesn't make sense to split them. I replaced it with a stage called FetchStage.

5. In the io module we can find the following files: [tile_pipelines.rs](https://github.com/maplibre/maplibre-rs/blob/main/maplibre/src/io/tile_pipelines.rs), [tile_repository.rs](https://github.com/maplibre/maplibre-rs/blob/main/maplibre/src/io/tile_repository.rs), [tile_request_state.rs](https://github.com/maplibre/maplibre-rs/blob/main/maplibre/src/io/tile_request_state.rs). In my opinion, these files are not really related to io and should move to a separate tile module.

6. I have made other small changes such as: using a style in headless mode to configure camera position and perspective, colors, layers...; use `while let Ok(result) = self.message_receiver.try_recv()` instead of `if let Ok(result) = self.message_receiver.try_recv()` to handle incoming mpsc messages.

This pull request attempts to fix the issues detailed above: simplify the diagram; remove the 2 mutexes; merge RequestStage and PopulateTileStoreStage; create a tile module; etc. I have updated my diagram with my new implementation below.

Screenshot 2022-06-19 at 11 35 27

**Please be aware that this is an opinionated refactor. I don't know every reasons behind the maplibre-rs implementation details and I may be wrong, or I may have misunderstood multiple aspects, such as why we use PipelineProcessor to process tiles? Is it for extendability?**

**I do not mind if we do not merge this pull request, but I believe we should discuss about simplification of the design and this might be a good place to start.**

## Additional issues detected

I have identified two other issues which I have not tackled yet.

1.  In RequestStage, when we send the styles (and a few other elements) to the thread, we use `.clone()` to make a copy of the style, so we can give the ownership to the thread. Instead of cloning, we should use `Arc<RwLock>` to provide an atomic reference counted access with the possibility to make multiple concurrent reads.
2. In RequestStage, we skip every layer that is not found within the `source-layer` property of the style layers. This may be wrong because layer styles that are not of type vector do not contain the `source-layer` property. [Source Layer property doc](https://docs.mapbox.com/help/glossary/source-layer/#:~:text=A%20source%20layer%20is%20an,data%20in%20a%20map%20style)

## ✔️ PR Todo

- [ ] Discuss the proposition
- [ ] Validate how GeometryIndex works
- [ ] Fix the issues with the headless mode demo
- [ ] Remove PipelineProcessor from the io module


maplibre/src/tile/tile_repository.rs
```diff
@@ -97,40 +114,44 @@ impl TileRepository {
...
+            cached_tile.status = TileStatus::Failed;
+            cached_tile.retry += 1;
+        }
+    }
 
-            return false;
+    /// Checks if a layer has been fetched.
+    pub fn needs_fetching(&self, coords: &WorldTileCoords) -> bool {
+        if let Some(cached_tile) = coords.build_quad_key().and_then(|key| self.tree.get(&key)) {
+            return cached_tile.status == TileStatus::Failed && cached_tile.retry < 3;
```
maxammann, line 154:
```suggestion
            return cached_tile.status == TileStatus::Failed && cached_tile.retry < 3;
```

The magic value 3 should be configurable.

maplibre/src/tile/mod.rs
```diff
@@ -0,0 +1,3 @@
+pub mod tile_parser;
```
maxammann, line 1: I would favor to name this module `data` instead of `tile`. That way it is more generic and is compatible with the naming in the paper.

maplibre-demo/src/main.rs
```diff
@@ -65,38 +61,49 @@ fn run_headless() {
...
+                        .map_context
+                        .tile_repository
+                        .put_tessellated_layer(TessellatedLayer {
+                            coords,
+                            buffer: vertex_buffer,
+                            feature_indices,
+                            layer_data: layer,
+                        });
+                }
+            }
```
maxammann, line 106: I think the previous definition (while not perfect) was simpler.

maplibre/src/tile/tile_repository.rs
```diff
@@ -37,15 +36,33 @@ impl StoredLayer {
     }
 }
 
+#[derive(PartialEq)]
+pub enum TileStatus {
+    Pending,
+    Failed,
+    Success,
+}
```
maxammann, line 44: :+1:

maplibre/src/stages/fetch_stage.rs
```diff
@@ -0,0 +1,216 @@
...
+    message_receiver: MessageReceiver,
+    scheduler: Scheduler<SM>,
+    http_source_client: HttpSourceClient<HC>,
+}
+
+impl<SM, HC> FetchStage<SM, HC>
+where
+    SM: ScheduleMethod,
+    HC: HttpClient,
+{
```
maxammann, line 33: I think its fine to merge the two stages :)

maplibre/src/stages/mod.rs
```diff
@@ -1,187 +1,19 @@
...
-    }
-}
-
-/// Stores and provides access to the thread safe data shared between the schedulers.
-#[derive(Clone)]
-pub struct SharedThreadState {
-    pub tile_request_state: Arc<Mutex<TileRequestState>>,
-    pub message_sender: mpsc::Sender<TessellateMessage>,
-    pub geometry_index: Arc<Mutex<GeometryIndex>>,
-}
```
maxammann, line 123: :fire:  We won't be be able to use Mutex on the web in the future because of COOP! So this is a very good change!

maxammann:
TODO: Reevaluate the Processor/DataPipeline

maxammann:
This will be split in at least 2 PRs:

* One for refactoring the tile repository
* One for refactoring the stages
* One for refactoring the pipelines

maxammann:
I opened this PR as a preparation: https://github.com/maplibre/maplibre-rs/pull/158

I noticed that the logic for loading new layers into the tile repository has been changed in this PR. Please create separate PRs for refactoring & changing actual logic.

Closing this as smaller PRs need to be created from this.

PR #143, opened 2022-07-04
Improve headless rendering

maxammann:
This changes a few things:

* Deduplicate code
* Actually make headless rendering working
* Move some files around
* Properly calculate camera height

### DIFF
--- a/maplibre-demo/src/main.rs
+++ b/maplibre-demo/src/main.rs
@@ -1,5 +1,5 @@
 use maplibre::benchmarking::tessellation::{IndexDataType, OverAlignedVertexBuffer};
-use maplibre::coords::{WorldTileCoords, ZoomLevel};
+use maplibre::coords::{TileCoords, ViewRegion, WorldTileCoords, ZoomLevel};
 use maplibre::error::Error;
 use maplibre::headless::HeadlessMapWindowConfig;
 use maplibre::io::pipeline::Processable;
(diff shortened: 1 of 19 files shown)


maplibre-demo/src/main.rs
```diff
@@ -117,5 +102,5 @@ fn main() {
     enable_tracing();
 
     run_headless();
-    run_in_window();
+    //run_in_window();
```
Drabble, line 105:
```suggestion
    run_in_window();
```

I think we can leave this one uncommented on the main branch.

maplibre/src/context.rs
```diff
@@ -12,9 +14,12 @@ pub struct ViewState {
...
 
 impl ViewState {
-    pub fn new(window_size: &WindowSize) -> Self {
+    pub fn new<P: Into<cgmath::Rad<f64>>>(window_size: &WindowSize, fovy: P) -> Self {
+        let center = TILE_SIZE / 2.0;
+        let fovy = fovy.into();
+        let height = center / (fovy / 2.0).tan();
         let camera = Camera::new(
-            (TILE_SIZE / 2.0, TILE_SIZE / 2.0, 150.0),
+            (center, center, height),
```
Drabble, line 22: I think we should still get the default values from the style configuration instead of showing the entire map. I can make that separatly in another PR if you agree?

Drabble (changes requested):
Overall it's all very good, I've tested it locally and it runs well. Only problem I see is the commented line in the demo, which should probably stay uncommented on the main branch.

maplibre/src/context.rs
maxammann: The height/fovy is not defined in the style.
The center position gets set in another PR.

PR #146, opened 2022-07-04
Update rust toolchain version

maxammann:
Update to 0.62.0 and latest nightly

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "1.60"
+channel = "1.62"
 targets = [
     "wasm32-unknown-unknown", "x86_64-unknown-linux-gnu", "x86_64-linux-android", "aarch64-linux-android",
     "x86_64-apple-darwin", "aarch64-apple-darwin", "x86_64-apple-ios", "aarch64-apple-ios", "aarch64-apple-ios-sim"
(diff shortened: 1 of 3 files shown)


rust-toolchain.toml
```diff
@@ -1,7 +1,7 @@
 [toolchain]
-channel = "1.60"
+channel = "1.62"
```
nyurik, line 2: Can this be set to "stable" instead?

nyurik (approved):

rust-toolchain.toml
maxammann: I think its nice to opt-into new versions. That way we we have always defined which minimal rust version is required to build the project.

In the future we maybe stick with older versions of rust (potentially).

nyurik: I would have kept it on stable until something breaks, and then switch to a fixed version if needed, but up to you

maxammann: We are also on a defined nightly version. Both should be updated at the same time. I think this way we have more control. But maybe we find some recommendation for this.

maxammann: Im not sure how cargo handles the following case:

A user has an old stable installed, then checks out the project, which requires a newer Rust version. Does it update the stable when `cargo build` is executed?

nyurik: Sure, your call :) I don't think cargo will download the latest on build.

PR #147, opened 2022-07-04
Limit the tiles in view

maxammann:
Previously we allowed infinitely many tiles in view. This caused buffer overflows.

This works around this issue:
```
thread 'main' panicked at 'Error in Queue::write_buffer: copy of 0..4216 would end up overrunning the bounds of the Destination buffer of size 2176'
```

### DIFF
--- a/maplibre/src/context.rs
+++ b/maplibre/src/context.rs
@@ -1,4 +1,4 @@
-use crate::coords::{Zoom, ZoomLevel, TILE_SIZE};
+use crate::coords::{ViewRegion, Zoom, ZoomLevel, TILE_SIZE};
 use crate::io::tile_repository::TileRepository;
 use crate::render::camera::{Camera, Perspective, ViewProjection};
 use crate::util::ChangeObserver;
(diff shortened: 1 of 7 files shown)


maxammann:
> This pull request does fix the issue with the buffer. However running this branch locally, I get a weird behaviour.

This is to be expected. In another PR I limit the pitch. This PR just limits the amount of tiles which can be shown.

Drabble:
> > This pull request does fix the issue with the buffer. However running this branch locally, I get a weird behaviour.
> 
> This is to be expected. In another PR I limit the pitch. This PR just limits the amount of tiles which can be shown.

Yes but my question is: why does the limit of tiles shown decreases as the pitch increase?

maxammann:
> > > This pull request does fix the issue with the buffer. However running this branch locally, I get a weird behaviour.
> > 
> > 
> > This is to be expected. In another PR I limit the pitch. This PR just limits the amount of tiles which can be shown.
> 
> Yes but my question is: why does the limit of tiles shown decreases as the pitch increase?

So by tilting more and more tiles get into view (Theoretically thats infinitely many).


Now:

Drabble (dismissed):

PR #149, opened 2022-07-04
Define a minimum and maximum pitch

maxammann:
In #147 we limited the maximium amount of tiles which can be loaded.
In this PR we limit the maximum pitch, else users hit a case in which tiles do not load even though they are in view.

The best value for the MAX_PITCH needs to be evaluated.


Drabble:
Do you think it would be possible to limit the pitch inside maplibre-winit instead (or to make it configurable)? I feel like it would give more freedom to the user. In some cases, I can imagine someone willing to use the map streetview-like. In that case, we should probably show a lower zoom tile for far away elements.

maxammann:
> Do you think it would be possible to limit the pitch inside maplibre-winit instead (or to make it configurable)? I feel like it would give more freedom to the user. In some cases, I can imagine someone willing to use the map streetview-like. In that case, we should probably show a lower zoom tile for far away elements.

I'd say this is a separate feature request. Feel free to open a feature request for this.
Ideally this would be configurable via a style specification I think.

PR #152, opened 2022-07-08
Update deps with cargo upgrade

birkskyum:
I noticed that quite many dependencies were behind their latest versions, and ran the cargo upgrade to get the project up to speed. All packages except of the ndk-glue for android are on their latest release.

### DIFF
--- a/maplibre-demo/Cargo.toml
+++ b/maplibre-demo/Cargo.toml
@@ -13,11 +13,11 @@ web-webgl = ["maplibre/web-webgl"]
 trace = ["maplibre/trace", "tracing-subscriber", "tracing-tracy", "tracy-client"]
 
 [dependencies]
-env_logger = "0.9"
+env_logger = "0.9.0"
 maplibre = { path = "../maplibre", version = "0.0.2", features = ["headless"]  }
 maplibre-winit = { path = "../maplibre-winit", version = "0.0.1"  }
 
-tracing = { version = "0.1" }
-tracing-subscriber = { version = "0.3", optional = true }
-tracing-tracy = { version = "0.8", optional = true }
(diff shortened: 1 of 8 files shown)


maplibre-demo/Cargo.toml
```diff
@@ -13,11 +13,11 @@ web-webgl = ["maplibre/web-webgl"]
...
 
-tracing = { version = "0.1" }
-tracing-subscriber = { version = "0.3", optional = true }
-tracing-tracy = { version = "0.8", optional = true }
-tracy-client = { version = "0.12.7", optional = true }
\ No newline at end of file
+tracing = "0.1.35"
+tracing-subscriber = { version = "0.3.14", optional = true }
+tracing-tracy = { version = "0.10.0", optional = true }
+tracy-client = { version = "0.14.0", optional = true }
```
maxammann, line 23: Tracy needs to stay on the version it was. The tracy version is linked to the tracy "inspector" version. We maybe can update the patch level slightly.

We should document this somehow.

maxammann (changes requested):

maplibre-demo/Cargo.toml
birkskyum: I reverted the update of the two tracy libs in both Cargo.toml so they stay in sync.

birkskyum: > We are on 0.7.2: https://github.com/nagisa/rust_tracy_client#version-support-table

Hmm... if we are on 0.7.2, isn't the current tracing-tracy=0.8 and tracy-client=0.12.7 too far ahead of the tracy? It seems like they line up at Tracy=0.7.8

PR #155, opened 2022-07-17
Upgrade to tokio 1.20

birkskyum:
https://github.com/tokio-rs/tokio/releases/tag/tokio-1.20.0


maxammann:
Are there any relevant changes we want? Else I'm not sure whether we want to accept such PRs.

birkskyum:
Ah sure, I must have missed that in the contribution guidelines - it was just maintenance as it didn't seem like there is any dependabot automation setup.

maxammann:
Are there actually contribution guidelines? :D 
I'm actually not sure what to do about updates. Generally we will need full-upgrades from time to time.

I think for any non-major updates we should update dependencies as needed.

PR #158, opened 2022-07-17
Introduce the tile module 

maxammann:
This should serve as a basis for upcoming PRs from #136


maxammann:
@Drabble can you go over this and check if we still want these changes?

maxammann:
Closing as there are too many conflicts now. I also merged some of the changes to main already.

PR #160, opened 2022-07-17
Add CLI to demo

maxammann:
This add a simple CLI for the demo. Can be extended in the future.

### DIFF
--- a/maplibre-demo/src/headless.rs
+++ b/maplibre-demo/src/headless.rs
@@ -1,31 +1,13 @@
-use std::collections::HashSet;
-
-use maplibre::coords::LatLon;
 use maplibre::{
-    benchmarking::tessellation::{IndexDataType, OverAlignedVertexBuffer},
-    coords::{TileCoords, ViewRegion, WorldTileCoords, ZoomLevel},
+    coords::{LatLon, WorldTileCoords},
     error::Error,
-    headless::{utils::HeadlessPipelineProcessor, HeadlessMapWindowConfig},
-    io::{
-        pipeline::{PipelineContext, PipelineProcessor, Processable},
-        source_client::{HttpClient, HttpSourceClient},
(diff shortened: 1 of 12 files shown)


birkskyum (approved):
Great with a headless mode. I find that it works as expected, and clap seems to be the only reasonable choice for this.

nyurik:
clap derive is awesome! I have been working with the author a bit, very knowledgeable and dedicated.  The structopt has merged into clap, so its a nobrained i guess :)

maplibre-demo/src/headless.rs
```diff
@@ -0,0 +1,55 @@
...
+        0,
+    );
+
+    for (z, x, y) in GridIterator::new(10, 10, tile_limits) {
+        let coords = WorldTileCoords::from((x as i32, y as i32, z.into()));
+        println!("Rendering {}", &coords);
+        map.map_schedule
+            .fetch_process(&coords)
+            .await
+            .expect("Failed to fetch and process!");
```
nyurik, line 44: should the "expect" report which tile it failed with?  Assuming this will increase cost only in the failed case.

maxammann:
> > I'm a little bit of a divergent opinion on the derive API. I basically found two issues while using it which make it a not so pleasant experience:
> > 
> > * There are runtime errors which theoretically could be caught during compile time. This goes slightly against Rusts idea: If it compiles it works!
> > * The derive API is not that nicely documented imo. I'm having a hard time to add for example a description for a subcommand.
> > 
> 
> Agree on the docs - might be good to spend some time helping the Clap author with it. WRT runtime errors - is there an issue for that? I would love to start discussing it, and hopefully address it too. I like Clap too much not to help it :)

I haven't looked deeper into it :D I though of it as expected behavior so i did not report it.
I just have a much better experience with the non-derive API.
Documentation is abolutely the first thing to improve I would say :D

Last time I was bruteforcing derive attributes :D

Issue #161, opened 2022-07-18
Limit the maximum and minimum zoom level


hanchao:
Keep zooming in or zooming out,  program will crashe. 

## 🤔 Expected Behavior

Need to limit maximum and minimum zoom level

## 😯 Current Behavior

Keep zooming in until the program crashe

'''
[2022-07-18T12:36:49Z INFO  maplibre::context] zoom: 30.42
[2022-07-18T12:36:49Z INFO  maplibre::context] zoom: 30.52
[2022-07-18T12:36:49Z INFO  maplibre::context] zoom: 30.71
[2022-07-18T12:36:49Z INFO  maplibre::context] zoom: 30.96
[2022-07-18T12:36:49Z INFO  maplibre::context] zoom: 31.21
[2022-07-18T12:36:49Z INFO  maplibre::context] zoom: 31.36
[2022-07-18T12:36:49Z INFO  maplibre::context] zoom: 31.53
[2022-07-18T12:36:49Z INFO  maplibre::context] zoom: 31.85
[2022-07-18T12:36:49Z INFO  maplibre::context] zoom: 32.16
thread 'main' panicked at 'attempt to add with overflow', maplibre/src/coords.rs:584:42
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[2022-07-18T12:36:49Z INFO  wgpu_core::hub] Dropping Global
[2022-07-18T12:36:49Z INFO  wgpu_core::device] Destroying 12 command encoders
'''

Keep zooming out until the program crashe

'''
[2022-07-18T12:35:05Z INFO  maplibre::context] zoom: -30.15
[2022-07-18T12:35:05Z INFO  maplibre::context] zoom: -30.41
[2022-07-18T12:35:05Z INFO  maplibre::context] zoom: -30.66
[2022-07-18T12:35:05Z INFO  maplibre::context] zoom: -30.9
[2022-07-18T12:35:05Z INFO  maplibre::context] zoom: -31.13
[2022-07-18T12:35:05Z INFO  maplibre::context] zoom: -31.35
[2022-07-18T12:35:05Z INFO  maplibre::context] zoom: -31.56
[2022-07-18T12:35:05Z INFO  maplibre::context] zoom: -31.76
[2022-07-18T12:35:05Z INFO  maplibre::context] zoom: -31.95
thread 'main' panicked at 'attempt to add with overflow', maplibre/src/coords.rs:584:42
[...]

Issue #73, opened 2022-05-03
Restrict navigation to map bounds


maxammann:
Currently, there are no bounds when navigating (pan, zoom) the map. The goal of this task is to restrict the navigation such that users can not escape and see empty space.

## 🤔 Expected Behavior

In Y direction the navigation should be restricted such that one can not pan out from the map.

In X direction it should be possible to infinitely pan. When panning beyond the bounds then the world should should be wrapped. This is done in ??.

## 😯 Current Behavior

There are no bounds.

## 💁 Possible Solution
??

MarkAnthonyM:
I'd love to take a stab at this and contribute, but I'm not really confident in my ability. Would it be possible to get a little mentorship if needed?

maxammann:
Sure thats my job :) feel free to join the matrix

DerKarlos:
As this seems a good and simple start, I had a look into it.
In coords.rs for Zoom I added
```
    pub fn limit(&mut self, min: f64, max: f64) {
        if        self.0 < min { self.0 = min }
        else { if self.0 > max { self.0 = max } }
    }
```
and in zoom_handler.rs in UpdateState for ZoomHandler I added
`next_zoom.limit(0.0, 20.0);`
The test runs well. Next I will look for panning / wrapping the world
What formal steps are needed, If I am done?

maxammann:
Hey @DerKarlos I think this is a good first task! Thanks for taking a look at it!

> and in zoom_handler.rs in UpdateState for ZoomHandler I added `next_zoom.limit(0.0, 20.0);`

That would temporarily allow to let the Zoom go above 20 or below 0. This might cause unexpected problems in the future. An implementation which does not use an additional `&mut self` function would be preferable I think.

Implementing this in the `ZoomHandler` is probably not what we want. The bounds will be a property of the style or map configuration in the `maplibre` crate. `ZoomHandler` is implemented in `maplibre-winit`. So the bounds/min/max values should be enforced in the `maplibre` core crate. Maybe a good place to implement this is the `ViewState` which holds the current zoom value.

DerKarlos:
I need time and help to get used to such complex projects like maplibre-rs.

**My 2nd atempt:**
I added min/maxzoom:f64 in Style&Default 0/20 (Mapstyle-Error if min>zoom or max<zoom)
and addet both into InteractiveMapSchedule/ViewState.
I impemented Zoom.clamp (no mod) and called it in ViewState::update_zoom.
[...]

PR #180, opened 2022-10-02
Limiting Zoom by style values

DerKarlos:
To solve the issue, I added a clamp to Zoom
and used it in ViewState.update_zoom.
I had to add maxzoom/minzoom as parameter/members to ViewState
Not yet changed: style-spec-v8.json
Todo: Error messages
(And it is my first Pull Request ever)

## 💻 Examples
Set the max/minzoom in the map style to the values, you want to test

## 🚨 Test instructions
Just zoom in/out to level 0/20. Zooming ends.
But the map sliding not! To be investigated

## ✔️ PR Todo

- [ ] Included links to related issues
#161 (not #73)


maxammann:
> style-spec-v8.json

That file should be removed soon. It is not used.

maplibre/src/context.rs
```diff
@@ -13,6 +13,8 @@ use crate::{
 /// Stores the camera configuration.
 pub struct ViewState {
     pub zoom: ChangeObserver<Zoom>,
+    pub maxzoom: u8,
```
maxammann, line 16:
```suggestion
    pub maxZoom: u8,
```

maplibre/src/style/style.rs
```diff
@@ -33,6 +35,8 @@ impl Default for Style {
             center: Some([46.5197, 6.6323]),
             pitch: Some(0.0),
             zoom: Some(13.0),
+            maxzoom: Some(20),
+            minzoom: Some( 0), // karlos: ok??? Detault values in extra file?
```
maxammann, line 39: Any values are fine here. This whole file is just a placeholder  basically until we have something more sophisticated.

maxammann, line 39: There should be an option in your IDE to run Rust fmt when saving the file.

maplibre/src/map_schedule.rs
```diff
@@ -43,13 +43,17 @@ impl<E: Environment> InteractiveMapSchedule<E> {
         wgpu_settings: WgpuSettings,
         renderer_settings: RendererSettings,
     ) -> Self {
+        let maxzoom = style.maxzoom.unwrap_or_default();
+        let minzoom = style.minzoom.unwrap_or_default();
+        // if maxzoom < zoom { error("mapstype error: maxzoom must be > zoom"); }
+        // if minzoom > zoom { error("mapstype error: minzoom must be < zoom"); }
```
maxammann, line 49: This brings up a great question: Where do we validate the style? There are probably more "invariants" we could check.
It should not happen in this file, because then we would need to copy&paste this code to the headless.rs file.

Ideally, it should be impossible to construct an invalid Style in the first place. Maybe we can add a constructor to the Style and validate it there.

For now it makes sense to a add a `self.validate()` function to the Style. We can call this then from anywhere.

maplibre/src/context.rs
```diff
@@ -13,6 +13,8 @@ use crate::{
 /// Stores the camera configuration.
 pub struct ViewState {
     pub zoom: ChangeObserver<Zoom>,
+    pub maxZoom: u8,
+    pub minzoom: u8,
```
maxammann, line 17: Same goes for this attribute. Also both should be f64. Zoom is a float. A zoom level is u8.

maxammann:
@DerKarlos I pushed a fix for the jumping. Basically the issue was that with every zoom there is also a translation of the map (such that the exact coordinates where the mouse points to, stays where the mouse pointer is)

maplibre/src/context.rs
DerKarlos: Also minzoom! But Rust requested, and now it is max_zoom
The symbols in all files mend to be identical? Also source.rs? There are a lot in style.rs!

maplibre/src/style/style.rs
DerKarlos: I do ussually, but your rustfmt.toml causes:
Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.
And with commenting out the toml, noting happens.

maplibre/src/map_schedule.rs
DerKarlos: I started, but far from perfect. Update soon

maplibre/src/context.rs
DerKarlos: Found the u8 in style.rs and used it. But yes, vector renderers never should force a "jumping" zoom.

DerKarlos:
Two editing the same thing may make trouble.
I could merge it but can't build:
maplibre-winit/src/input/zoom_handler.rs:22:56
   |
22 |                 let zoom_changed = actual_next_zoom != current_zoom;
   |                                                        ^^^^^^^^^^^^ expected `()`, found struct `Zoom`

maplibre/src/style/style.rs
maxammann: Oh yes, you will need nightly fmt. So you will continue to use stable Rust, but you can use nightly formatter.

Try to run: just format or just fmt

DerKarlos: Not found: just. Is a windows package manager? I use macOS.
rustup install nightly-x86_64-apple-darwin
But still: rustfmt 1.4.38-stable (e092d0b6 2022-07-16)
% rustup component add rustfmt --toolchain nightly
info: component 'rustfmt' for target 'x86_64-apple-darwin' is up to date  :(  no nightly-x86

maxammann: In that case try to run cargo +nightly fmt

You probably have it already installed :)

DerKarlos: Yes! So I have to add "+nightly". I will try to change this in the VS-Code

maxammann:
@DerKarlos merged in main to fix builds

maxammann:
@DerKarlos Hi! Are you still interested in pushing this forward? I may be able to help resolving conflicts.

maxammann:
@DerKarlos Still wanna merge this? I think this is close to being finished

DerKarlos:
Sorry, live put me to other priorities. You may finish it or delete it.
May be I will be back some day ...

PR #192, opened 2022-11-01
Implement 3d extrusion of buildings

Drabble:
This is a simple implementation for showcasing 3d possibilities in Maplibre-rs. 

We simply take the elements in the buildings layer and extrude them with a fixed height.

In the future, it would be very interesting to add details to the buildings that are extruded and extrude based on the height metadata available in OSM.

Extrusion can be enabled/disabled in the `Style` configuration.

Finally I probably should split the function in `tile_pipelines.rs` into smaller functions to make it easier to understand.

### DIFF
--- a/maplibre/src/render/shaders/mod.rs
+++ b/maplibre/src/render/shaders/mod.rs
@@ -99,13 +99,13 @@ impl Shader for TileShader {
                         // position
                         wgpu::VertexAttribute {
                             offset: 0,
-                            format: wgpu::VertexFormat::Float32x2,
+                            format: wgpu::VertexFormat::Float32x3,
                             shader_location: 0,
                         },
                         // normal
                         wgpu::VertexAttribute {
-                            offset: wgpu::VertexFormat::Float32x2.size(),
-                            format: wgpu::VertexFormat::Float32x2,
+                            offset: wgpu::VertexFormat::Float32x3.size(),
+                            format: wgpu::VertexFormat::Float32x3,
                             shader_location: 1,
                         },
                     ],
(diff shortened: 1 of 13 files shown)


maxammann:
Converted to draft, or it this ready for review? :)

Drabble:
> Converted to draft, or it this ready for review? :)

I think it's not quite ready for production but being an experimental functionality, we could just disable it by default and enable it when we want to.

I guess the main thing left to do is to cleanup the big function with the extrusion to simplify the code and review/discuss about the implementation details.

DerKarlos:
Yes please, 3D buildings! The changes are not that much:
The renderer gets an optional 3D extrude. Therefore some APIs are extended form 2D to 3D.
Style access and normals are needed, why a camera to?
The extrude code is quite plain. Its good to see, how the the faces are build.
I would like to use/test it! How could I merge it into my fork?
I compared it with the 3D renderer OSM2World. We should think about it's concepts.
I could introduce, how it works if you like and we could discus a merge of ideas with maplibre-rs.
https://wiki.openstreetmap.org/wiki/OSM2World
We could create an FRC for 3D object rendering

maxammann:
I'll resolve merge conflicts now!

maplibre/src/render/shaders/mod.rs
```diff
@@ -99,13 +99,13 @@ impl Shader for TileShader {
                         // position
                         wgpu::VertexAttribute {
                             offset: 0,
-                            format: wgpu::VertexFormat::Float32x2,
+                            format: wgpu::VertexFormat::Float32x3,
```
maxammann, line 102: I'm not sure if we should upgrade from 2D to 3D for everything.
This will increase the required bandwidth to the GPU.

But I'm not sure how much the impact is performance wise.

maxammann:
@Drabble  Hi! Are you still interested in pushing this forward? :) I guess a lot changed, but I think it would be nice to keep this effort updated.

PR #199, opened 2022-11-03
Postponed: Self-hosted CI runner for macOS

maxammann:
We need to clarify security requirements: https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security

### DIFF
--- a/.github/workflows/library-apple.yml
+++ b/.github/workflows/library-apple.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   library-apple:
     name: Build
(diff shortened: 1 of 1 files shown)


maxammann:
Requires: [maplibre/maplibre#](https://github.com/maplibre/maplibre/issues/142)

maxammann:
It's unlikely that this will happen anytime soon. Also because self-hosted runners are hard to setup.

PR #200, opened 2022-11-03
Fix demo startup without Vulkan

vollkorntomate:
A bug was introduced in commit fdd09e0 which led to the demo application not starting on macOS (Apple Silicon) and possibly other systems that don't use Vulkan as GPU backend.

This PR resolves this issue.

Previous error:
```
thread 'main' panicked at 'Unable to find a GPU! Make sure you have installed required drivers!',  .../maplibre-rs/maplibre/src/render/mod.rs:249:14
```

I couldn't test it on other systems, but I'm quite certain that this change doesn't break it there.

### DIFF
--- a/android/Cargo.toml
+++ b/android/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2021"
 publish = false
 
 [dependencies]
-maplibre = { path = "../maplibre" }
+maplibre = { path = "../maplibre", features = ["thread-safe-futures"] }
 maplibre-winit = { path = "../maplibre-winit", version = "0.0.1"  }
 env_logger = "0.9.0"
 log = "0.4.17"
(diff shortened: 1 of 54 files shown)


maxammann:
Hi @vollkorntomate, thanks for your PR!

There is actually another usage of `Backends::VULKAN` in the maplibre-wgpu module. It is located in the EventLoop impl in the lib.rs file.

On Android an late initialization of the renderer is required. This means that during the event loop the same functionality is required.
Somehow we will need to make either the whole `RendererBuilder` or the `RendererSettings` + `WgpuSettings` available in the `Map` struct. If it is presisted in the `Map` struct then we can access it during the event loop and use the correct init params.

So in order to merge this is would be great if you can:
* either replace both occurences
* Or use the same renderer init parameters in the event loop as in the function you already touched.

maxammann:
Maybe we can also replace the rendering init with a closure which holds the rendering settings.

PR #204, opened 2022-11-06
Add transferable-memory create to fix regressions

maxammann:
Fixes regressions from #174

For example we had now a max. byte length for vertices and indices.


maxammann:
Closing in favor of #221

Issue #71, opened 2022-05-03
Raster Tile Rendering


maxammann:
We want to be able to render raster tiles. This is required for rendering data sets like "Natural Earth". Raster tiles can be of any web-native image format like PNG or WEBP.

## 🤔 Expected Behavior

Raster tiles need to be scaled and transformed like vector tiles.

## 😯 Current Behavior

No raster tiles are supported.

## 💁 Possible Solution

The WebGPU spec offers a way to load bitmaps to textures: https://toji.github.io/webgpu-best-practices/img-textures.html
In the browser this requires the `createImageBitmap()` API which offloads decoding of images to the browser.

ybiletskyi:
Any ideas about implementation of this feature?

ybiletskyi:
Got it. Thanks for update. I'll try to implement it in wgpu.

quietlychris:
Just as a note: I have a personal interest in this feature! At the moment, I'm using MapLibre (JavaScript) on my personal [site](https://cmoran.xyz/geospatial) to overlay .webp images of environmental drone imagery I've collected. The site's backend is actually written in Rust, and I'd love the opportunity to re-write this particular application with maplibre-rs as well, which is dependent on raster imagery being available (and symbols). Really excited about this project, so just wanted to mention a use case I'd personally be delighted to have support for at some point. I don't have a ton of capacity for new contributions at the moment, but if you need support for something like testing for either this project or the wgpu feature, just let me know!

ybiletskyi:
Hey. I just created a [PR](https://github.com/gfx-rs/wgpu/pull/2781) to wgpu. It probably take a time to merge it. But it's not a blocker any more for start this feature implementation.

maxammann:
@ybiletskyi, @Quillasp might also be interested in working on this as part of his bachelor thesis. I wondered whether collaboration could make sense. If not it shouldn't be hard to find two distinct areas to work on in maplibre-rs.

maxammann:
@Drabble can you check with @Quillasp?

Quillasp:
@Drabble gave me an overview of the project today. But the project is still quite big (for me at least), do you have any idea where I should start looking in the code to see where I could implement the basics of this feature ? (sorry for the noob question)

ybiletskyi:
[...]

PR #205, opened 2022-11-11
Raster Tiles

Quillasp:
Associated with the issue #71. 

This draft PR can load and display the raster tiles. It still needs a way to do different requests based on the tile type (vector or raster for the moment).

## ✔️ PR Todo

- [x] Can do both types of requests
- [x] Raster data loading and storing
- [x] Raster rendering


maxammann:
Just tried to run it on Linux. It compiles and runs, but its glitching quite a bit.
Do you think that is a bug with my environment or another issue?

Quillasp:
> Just tried to run it on Linux. It compiles and runs, but its glitching quite a bit. Do you think that is a bug with my environment or another issue?

I don't have a Linux machine so I can't test right away but it shouldn't be different from other platforms. 🤔

maxammann:
@Quillasp web is failing. We probably miss some more feature flags

maplibre/src/render/render_commands.rs
```diff
@@ -71,6 +92,22 @@ impl<P: PhaseItem> RenderCommand<P> for SetTilePipeline {
...
+    ) -> RenderCommandResult {
+        if let Initialized(raster_resources) = &state.raster_resources {
+            pass.set_render_pipeline(raster_resources.pipeline.as_ref().unwrap());
+            RenderCommandResult::Success
+        } else {
+            RenderCommandResult::Failure
+        }
+    }
+}
+
```
maxammann, line 110: Can we somehow reuse the code from above? it is identical

maplibre/src/render/resource/buffer_pool.rs
```diff
@@ -271,34 +274,76 @@ impl<Q: Queue<B>, B, V: Pod, I: Pod, TM: Pod, FM: Pod> BufferPool<Q, B, V, I, TM
 
         queue.write_buffer(
             &self.layer_metadata.inner,
-            maybe_entry.buffer_layer_metadata.start,
+            maybe_entry.buffer_layer_metadata.as_ref().unwrap().start,
```
maxammann, line 277: These unwraps are probably bad

maplibre/src/render/resource/buffer_pool.rs
```diff
@@ -446,12 +496,14 @@ pub struct IndexEntry {
...
-    buffer_layer_metadata: Range<wgpu::BufferAddress>,
+    buffer_layer_metadata: Option<Range<wgpu::BufferAddress>>,
     // Range of bytes within the backing buffer for feature metadata
-    buffer_feature_metadata: Range<wgpu::BufferAddress>,
+    buffer_feature_metadata: Option<Range<wgpu::BufferAddress>>,
     // Amount of actually usable indices. Each index has the size/format `IndexDataType`.
     // Can be lower than size(buffer_indices) / indices_stride because of alignment.
     usable_indices: u32,
+    // Indicate if current entry represent raster tile
+    raster: bool,
```
maxammann, line 506: I think we should not mix the raster tiles into this buffer pool.

The buffer pool is intended to be used when you have thousand of vertices. It can be used to provide views into the buffer and automatically discarding older entires

maxammann: So we need to come up with some abstraction

maplibre/src/render/resource/buffer_pool.rs
```diff
@@ -612,6 +663,16 @@ mod tests {
...
+
+        fn write_texture(
+            &self,
+            texture: ImageCopyTexture,
+            data: &[u8],
+            data_layout: ImageDataLayout,
+            size: Extent3d,
+        ) {
+            todo!()
+        }
```
maxammann, line 675: Should be removed probably

maplibre/src/render/resource/raster.rs
```diff
@@ -0,0 +1,142 @@
...
+        usage: wgpu::TextureUsages,
+    ) {
+        self.texture = Some(Texture::new(
+            label,
+            device,
+            format,
+            width,
+            height,
+            self.msaa.unwrap().clone(),
+            usage,
```
maxammann, line 70: Maybe fall back to some default

maplibre/src/render/shaders/tile_raster.vertex.wgsl
```diff
@@ -0,0 +1,24 @@
...
+    @location(9) zoom_factor: f32,
+) -> VertexOutput {
+    let z = 0.0;
+    let width = 3.0 * zoom_factor;
+    let normal = vec2<f32>(0.0,0.0);
+
+    var position = mat4x4<f32>(translate1, translate2, translate3, translate4) * vec4<f32>(position + normal * width, z, 1.0);
+    position.z = 1.0;
+
+    return VertexOutput(tex_coords, position);
```
maxammann, line 23: Maybe some stuff is duplicated here. Fine for now though!

maplibre/src/render/stages/upload_stage.rs
```diff
@@ -223,6 +238,66 @@ impl UploadStage {
...
+                                                        4 * width,
+                                                    ),
+                                                    rows_per_image: std::num::NonZeroU32::new(
+                                                        height,
+                                                    ),
+                                                },
+                                                raster_resources
+                                                    .texture
+                                                    .as_ref()
+                                                    .unwrap()
```
maxammann, line 292: If possible remove that unwrap

maplibre/src/render/stages/upload_stage.rs
```diff
@@ -223,6 +238,66 @@ impl UploadStage {
                                         &feature_metadata,
                                     );
                                 }
+                                StoredLayer::RasterLayer {
```
maxammann, line 241: I would favor to create a new stage: upload_raster_stage and then conditionally add stat stage, instead of having the inline cfg flag

maplibre/src/render/tile_pipeline.rs
```diff
@@ -84,6 +87,25 @@ impl RenderPipeline for TilePipeline {
                     },
                     count: None,
                 }]])
+            } else if self.raster {
```
maxammann, line 90: Not sure yet how we can improve this. This is very specific for a file which tries to keep generic.

maplibre/src/render/stages/upload_stage.rs
```diff
@@ -223,6 +238,66 @@ impl UploadStage {
...
+                                    #[cfg(feature = "raster")]
+                                    {
+                                        if let Initialized(raster_resources) = raster_resources {
+                                            buffer_pool.allocate_layer_raster(
+                                                queue,
+                                                *coords,
+                                                style_layer.clone(),
+                                            );
+                                            let img = image::load_from_memory(&layer_data).unwrap();
+                                            let rgba = img.to_rgba8();
```
maxammann, line 255: We are here in the render loop. Decoding of images should happen in the pipeline and not in the loop.

maxammann (changes requested):
Looks very cool already! I mean yeah there are a few rough edges, but that is also due to some architectural changes we need to do in order to add the feature.

maplibre/src/stages/request_stage.rs
```diff
@@ -77,7 +83,12 @@ pub fn schedule<
         let coords = input.coords;
         let client = context.source_client();
 
-        match client.fetch(&coords).await {
+        #[cfg(feature = "raster")]
+        let source = SourceType::Raster(RasterSource::default());
+        #[cfg(not(feature = "raster"))]
+        let source = SourceType::Tessellate(TessellateSource::default());
```
maxammann, line 89: Does that mean if raster is enabled then vector tiles do no longer work?

maxammann:
I introduced quite a lot of merge conflicts. I will resolve them today!

maxammann:
Done, I also simplified the pipeline slightly and updated the APC implementation for Web.

maxammann:
@Quillasp I pushed a simplification which does not require the BufferPool

maxammann:
Merged in https://github.com/maplibre/maplibre-rs/pull/253

Issue #201, opened 2022-11-05
Upgrade Rust to 1.65


maxammann:
Rust 1.65 was recently released. Update it and the nightly version. The version of the nightly version should be that of today.

* Update version in rust-toolchain.toml
* Update version in Justfile
* Update version in NDK project

Do a quick search over the project for the Rust versions.

maxammann:
done in #207

PR #206, opened 2022-11-11
Clippy fixes

julienr:
Fixes a number of clippy issues:
- Implement `Default` for various types
- Remove some useless format/mut
- Simplify some `if let match`
- Fix some cases where variables are never read because they are immediatly overwritten

### DIFF
--- a/maplibre/src/io/geometry_index.rs
+++ b/maplibre/src/io/geometry_index.rs
@@ -62,6 +62,12 @@ impl GeometryIndex {
     }
 }
 
+impl Default for GeometryIndex {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 /// Index of tiles which can be of two types: spatial or linear.
 /// Spatial tiles are stored in a multi-dimentional tree which represents their position in the tile.
 /// Linear tiles are simply stored in a vector.
@@ -195,6 +201,12 @@ impl IndexProcessor {
 
     pub fn get_geometries(self) -> Vec<IndexedGeometry<f64>> {
         self.geometries
+    }
+}
+
+impl Default for IndexProcessor {
+    fn default() -> Self {
+        Self::new()
     }
 }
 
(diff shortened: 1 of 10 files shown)


nyurik (approved):
lgtm

julienr:
Looks like the build is failing because some dependencies require rust 1.65:
```
Caused by:
  package `wgpu-hal v0.14.0 (/home/runner/.cargo/git/checkouts/wgpu-53e70f8674b08dd4/d536613/wgpu-hal)` cannot be built because it requires rustc 1.65 or newer, while the currently active rustc version is 1.64.0
```
I guessed this is blocked by #201 then ?

nyurik:
@julienr :shrug: I think @maxammann would have to answer that. Or you could try to tackle #201?

maxammann:
Its a new issue :) Ill fix it today propably.

maxammann:
I'll merge as soon as checks succeed: https://github.com/maplibre/maplibre-rs/pull/207

maplibre/src/io/geometry_index.rs
```diff
@@ -62,6 +62,12 @@ impl GeometryIndex {
     }
 }
 
+impl Default for GeometryIndex {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
maxammann, line 69: I was wondering which was is better: 
* Implement the constructor in new()
* or implement it in default() and then call default() in new()

What do you think?

maxammann (approved):

maplibre/src/io/geometry_index.rs
julienr: No strong feeling and I'm not a Rust expert. But in the [clippy docs](https://rust-lang.github.io/rust-clippy/master/#new_without_default) they seem to point ot this pattern (just have default delegate to new)

maxammann: Oke cool, that let's use that style!

PR #208, opened 2022-11-12
justfile: automatically install wasm32 target for nightly toolchain

julienr:
This fixes an issue where `just web-lib build` would fail because it cannot find the 'core' crate.

I am not sure if this is the right way to fix the problem or the right place, but it worked for me.

## 💻 Examples

```
$ just web-lib build
...
Creating WASM...
   Compiling proc-macro2 v1.0.47
   Compiling unicode-ident v1.0.5
   Compiling quote v1.0.21
   Compiling syn v1.0.103
   Compiling autocfg v1.1.0
   Compiling log v0.4.17
   Compiling cfg-if v1.0.0
error[E0463]: can't find crate for `core`
  |
  = note: the `wasm32-unknown-unknown` target may not be installed
  = help: consider downloading the target with `rustup target add wasm32-unknown-unknown`
  = help: consider building the standard library from source with `cargo build -Zbuild-std`
```

## 🚨 Test instructions

- Check the repo without any nightly rust toolchain installed on Ubuntu 22.04
- Run `just web-lib build`

## ✔️ PR Todo

- [ ] Decide if this is the right place to install this target
- [ ] Should we do the same for the stable toolchain ?

### DIFF
--- a/justfile
+++ b/justfile
@@ -104,14 +104,16 @@ web-test FEATURES: nightly-toolchain (nightly-targets "wasm32-unknown-unknown")
 
 build-android: build-android-lib build-android-demo
 
-build-android-lib: nightly-toolchain print-android-env
-  export RUSTUP_TOOLCHAIN=$NIGHTLY_TOOLCHAIN && cd android/gradle && ./gradlew :lib:assembleDebug
+ensure-android-toolchain: nightly-toolchain (nightly-targets "x86_64-linux-android" "aarch64-linux-android" "i686-linux-android")
 
-build-android-demo: nightly-toolchain print-android-env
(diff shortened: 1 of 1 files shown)


maxammann:
Hey @julienr thanks for your contribution! I can definitely also see the problem. In the CI I we manually install the target. What do you think about adding the task dependencies? 

I commited to your branch.

julienr:
You way looks better indeed !

maxammann:
> Your way looks better indeed !

I'd say we can merge this as is. I did not yet do the changes for apple or android.

maxammann (approved):

PR #210, opened 2022-11-12
Add MSVC setup to CI

maxammann:
This setups the MSVC compiler in our CI. Right now we do not need it.

But as soon as we link for example against SQLite or OpenSSL we might need it again.

## ✔️ PR Todo

### DIFF
--- a/.github/workflows/demo-windows.yml
+++ b/.github/workflows/demo-windows.yml
@@ -14,6 +14,33 @@ jobs:
       - name: Install mysys2 dependencies
         shell: powershell
         run: pacman -S --noconfirm unzip
+      - name: Setup MSVC
+        run: |
+          $script=@'
+          Push-Location "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build"
+          & "C:\Windows\System32\cmd.exe" /c "vcvars64.bat & set" |
+          ForEach-Object {
+            if ($_ -match "=") {
+              $v = $_.split("=", 2)
+              echo "$v[0]=$v[1]" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf8 -Append
+            }
+          }
+          Pop-Location
+          '@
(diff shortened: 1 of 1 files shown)


.github/workflows/demo-windows.yml
```diff
@@ -8,9 +8,38 @@ jobs:
...
+        run: |
+          $script=@'
+          Push-Location "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build"
+          & "C:\Windows\System32\cmd.exe" /c "vcvars64.bat & set" |
+          ForEach-Object {
+            if ($_ -match "=") {
+              $v = $_.split("=", 2)
+          
+              if ($($v[0]) -match "Path") {
+                echo "$($v[0])=$($v[1]);C:\msys64\usr\bin" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf8 -Append
```
maxammann, line 21: Cleanup this case. We are adding msys64 at the very last in the `Path` such that binaries like `link.exe` have a higher priority

.github/workflows/demo-windows.yml
```diff
@@ -8,15 +8,47 @@ jobs:
...
       - name: Install mysys2 dependencies
         shell: powershell
         run: pacman -S --noconfirm unzip
-      - name: Show PATH
+      - name: Show Powershell PATH
         shell: powershell
         run: Write-Host $env:path
+      - name: Show bash PATH
+        shell: powershell
+        run: bash -c 'echo $PATH'
```
maxammann, line 51: TODO: Currently the git bash is executed here. The Git bash also has a `link.exe`. Make sure that either Git bash or msys2 bash both prefer using the `link.exe` from MSVC.

maxammann: Idea: Use powershell to write to .bashrc and overwrite at least `PATH` variable.

PR #220, opened 2022-11-20
Improve tile view pattern

maxammann:
I found these commits on my local machine. Right now I'm not exactly sure what they are doing


maxammann:
Change are no longer relevant, diagram was moved to #250

PR #222, opened 2022-12-11
Improve error handling

maxammann:
Redesigned error types across the project according to: https://mmapped.blog/posts/12-rust-error-handling.html


maplibre/src/io/apc.rs
```diff
@@ -41,18 +39,67 @@ pub enum Input {
     NotYetImplemented, // TODO: Placeholder, should be removed when second input is added
 }
 
+#[derive(Debug)]
+pub enum SendError {
+    // Error occured during transmission
+    Transmission,
+}
+
+impl Display for SendError {
```
nyurik, line 48: this seems weird, but works :)   Do you really need a Display trait on an error (here and in another trait)

web/src/platform/singlethreaded/wasm_entries.rs
```diff
@@ -1,95 +1,89 @@
...
     let context = PassingContext {
         source_client: SourceClient::new(HttpSourceClient::new(WHATWGFetchHttpClient::new())),
     };
 
-    let result = (procedure)(input, context).await;
-
-    if let Err(e) = result {
-        error!("{:?}", e); // TODO handle better
-    }
+    (procedure)(input, context).await?;
```
nyurik, line 42: just for personal education - why do you need to put a var in the parenthesis here?

nyurik (approved):
lgtm, a few minor questions

web/src/platform/singlethreaded/wasm_entries.rs
maxammann: They are probably not required here. Though there are cases where you need them. I think if `procedure` is part of a struct, then you have to do `(structure.procedure)()` because you are calling the lambda and not a function defined on the struct.

A new opportunity for a clippy lint? :D

maplibre/src/io/apc.rs
maxammann: I will replace all manual implementations with thiserror derives: https://crates.io/crates/thiserror

maxammann: Doing this also in this PR

nyurik (approved):

PR #233, opened 2022-12-22
Transform improvements

maxammann:
~~Implements perspective transforms like in maplibre-gl-js~~

This fundamentally reworks how the far distance plane is calculated. First I wanted to implement it like in maplibre-gl-js. Though that appraoch did not allow yaw AND pitch in both directions. It only allowed pitch in a single direction (only 90°, not the full 180°).

The new approach is dicussed here: https://gamedev.stackexchange.com/questions/207328/calculation-of-far-distance-plane-based-on-yaw-and-pitch-for-a-map-renderer

It works in all directions and should also be way faster to calculate.

This PR also fixes the amount of tiles that are whown for the current zoom. If calculates the appropriate zoom level based on the current zoom and takes into account the resolution of tiles.

### DIFF
--- a/maplibre-winit/src/input/camera_handler.rs
+++ b/maplibre-winit/src/input/camera_handler.rs
@@ -5,27 +5,39 @@ use maplibre::context::MapContext;
 
 use super::UpdateState;
 
-pub struct TiltHandler {
+pub struct CameraHandler {
     delta_pitch: Deg<f64>,
+    delta_roll: Deg<f64>,
+    delta_yaw: Deg<f64>,
 
     speed: f64,
     sensitivity: f64,
 }
 
-impl UpdateState for TiltHandler {
+impl UpdateState for CameraHandler {
     fn update_state(&mut self, MapContext { view_state, .. }: &mut MapContext, dt: Duration) {
         let dt = dt.as_secs_f64() * (1.0 / self.speed);
 
-        let delta = self.delta_pitch * dt;
-        view_state.camera_mut().tilt(delta);
-        self.delta_pitch -= delta;
+        let delta_pitch = self.delta_pitch * dt;
+        view_state.camera_mut().pitch(delta_pitch);
+        self.delta_pitch -= delta_pitch;
+
+        let delta_roll = self.delta_roll * dt;
+        view_state.camera_mut().roll(delta_roll);
+        self.delta_roll -= delta_roll;
+
+        let delta_yaw = self.delta_yaw * dt;
+        view_state.camera_mut().yaw(delta_yaw);
+        self.delta_yaw -= delta_yaw;
     }
 }
 
-impl TiltHandler {
+impl CameraHandler {
     pub fn new(speed: f64, sensitivity: f64) -> Self {
         Self {
             delta_pitch: Deg::zero(),
+            delta_roll: Deg::zero(),
+            delta_yaw: Deg::zero(),
             speed,
             sensitivity,
         }
@@ -48,6 +60,22 @@ impl TiltHandler {
             }
             winit::event::VirtualKeyCode::F => {
                 self.delta_pitch += amount;
+                true
+            }
+            winit::event::VirtualKeyCode::T => {
+                self.delta_yaw -= amount;
+                true
+            }
+            winit::event::VirtualKeyCode::G => {
+                self.delta_yaw += amount;
+                true
+            }
+            winit::event::VirtualKeyCode::Z => {
+                self.delta_roll -= amount;
+                true
+            }
+            winit::event::VirtualKeyCode::H => {
+                self.delta_roll += amount;
                 true
             }
             _ => false,
--- a/maplibre-winit/src/input/camera_handler.rs
+++ b/maplibre-winit/src/input/camera_handler.rs
@@ -1,84 +1,107 @@
 use std::time::Duration;
 
-use cgmath::{Deg, Zero};
+use cgmath::{Deg, MetricSpace, Rad, Vector2};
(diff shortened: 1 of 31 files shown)


nyurik (dismissed):
lgtm

maplibre/src/vector/upload_system.rs
```diff
@@ -32,7 +33,8 @@ pub fn upload_system(
         &mut Eventually<VectorBufferPool>,
     >() else { return; };
 
-    let view_region = view_state.create_view_region();
+    let view_region =
+        view_state.create_view_region(view_state.zoom().zoom_level(DEFAULT_TILE_SIZE));
```
nyurik, line 37: this looks weird name-wise -- the `zoom_level` implies "zoom to this zoom level", but the parameter is "tile size", not some zoom.

maxammann: It is a getter. Should be renamed maybe.

nyurik: `get_zoom()` ?

nyurik:
soooooooo happy to see you hacking away on this :)

maplibre-winit/src/input/mod.rs
```diff
@@ -77,13 +83,17 @@ impl InputController {
                 ..
             } => {
                 if !self.shift_handler.process_key_press(*key, *state) {
-                    if !self.tilt_handler.process_key_press(*key, *state) {
-                        self.zoom_handler.process_key_press(*key, *state)
+                    if !self.debug_handler.process_key_press(*key, *state) {
```
nyurik, line 86: why not replace the whole thing with an OR condition?  I think this is identical:

```rust
self.shift_handler.process_key_press(*key, *state) ||
self.debug_handler.process_key_press(*key, *state) ||
self.zoom_handler.process_key_press(*key, *state)
```

maplibre-winit/src/input/mod.rs
```diff
@@ -107,6 +117,8 @@ impl InputController {
                         .process_window_position(&Vector2::from(position), true);
                     self.zoom_handler
                         .process_window_position(&Vector2::from(position), true);
+                    self.camera_handler
+                        .process_window_position(&Vector2::from(position), true);
```
nyurik, line 121: do you need to redo `Vector2::from(position)` once for each call?

maplibre-winit/src/input/zoom_handler.rs
```diff
@@ -24,18 +24,15 @@ impl UpdateState for ZoomHandler {
                 let view_proj = view_state.view_projection();
                 let inverted_view_proj = view_proj.invert();
 
-                if let Some(cursor_position) = view_state.camera().window_to_world_at_ground(
+                if let Some(cursor_position) = view_state.window_to_world_at_ground(
```
nyurik, line 27: try using slightly shorter names for vars when that var is used just for a few lines, and it is relatively easy to understand the meaning, e.g. `cursor_pos`?  I tend to use a single character vars for oneliners as those are the easiest

maplibre/src/raster/request_system.rs
```diff
@@ -48,7 +49,8 @@ impl<E: Environment, T: RasterTransferables> System for RequestSystem<E, T> {
         }: &mut MapContext,
     ) {
         let _tiles = &mut world.tiles;
-        let view_region = view_state.create_view_region();
+        let view_region =
+            view_state.create_view_region(view_state.zoom().zoom_level(DEFAULT_TILE_SIZE));
```
nyurik, line 53: not certain of the reasons -- the `.zoom().zoom_level(...)` seems weird... any way to make it into one call, like `.zoom_level(...)`?

nyurik (commented):
a few minor suggestions

maplibre-winit/src/input/camera_handler.rs
```diff
@@ -0,0 +1,107 @@
...
+                    .into();
+                let previous = *self.start_delta_pitch.get_or_insert(camera.get_pitch());
+                camera.set_pitch(previous + delta);
+            }
+        }
+    }
+}
+
+impl CameraHandler {
+    pub fn new(sensitivity: f64) -> Self {
```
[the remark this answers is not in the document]
maxammann: I don't think that is a good idea because a sensitivity of 0 does not make sense.

maplibre-winit/src/input/zoom_handler.rs
maxammann: I don't want to touch these functions right now because they contain heavy maths and the name makes sense from that code's perspective.

maplibre/src/raster/request_system.rs
maxammann: I could add a shorthand, not sure if that is worth it though. That call gives the zoom_level at a specific zoom. So the call makes sense even though it looks weird.

maxammann:
> soooooooo happy to see you hacking away on this :)

Thanks! :) Yeah need to get this tidied up for EuroRust :D

maxammann:
These apple executors are barely available :(

maplibre-winit/src/input/mod.rs
nyurik: might make sense to push the changes before solving the conv - easier to keep track of what was done/rejected vs pending

maplibre-winit/src/input/camera_handler.rs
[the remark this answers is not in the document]
maxammann: 👍🏼  I'll do that in some other PR

PR #235, opened 2023-01-08
Bump prost crate version to 0.11.5

Tristramg:
The previous version (0.10.0) wouldn’t compile (rust 1.65, `cargo run -p maplibre-demo`, linux wayland) with the following error:

```
error[E0599]: no function or associated item named `decode` found for struct `Tile` in the current scope
  --> maplibre/src/io/tile_pipelines.rs:28:40
   |
28 |         let tile = geozero::mvt::Tile::decode(data.as_ref()).expect("failed to load tile");
   |                                        ^^^^^^ function or associated item not found in `Tile`
   |
   = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
   |
1  | use prost::message::Message;
   |

error: unused import: `prost::Message`
  --> maplibre/src/io/tile_pipelines.rs:4:5
   |
4  | use prost::Message;
   |     ^^^^^^^^^^^^^^
   |
```

### DIFF
--- a/.github/actions/install-system-dependencies/action.yml
+++ b/.github/actions/install-system-dependencies/action.yml
@@ -15,7 +15,7 @@ runs:
         sudo apt-get install -y \
           protobuf-compiler \
           libwayland-dev libxkbcommon-dev # Required for winit
-    - name: Install System Dependencies (Ubuntu)
+    - name: Install System Dependencies (macOS)
       if: runner.os == 'macOS'
       shell: bash
       run: brew install protobuf
--- a/.github/workflows/library-apple.yml
+++ b/.github/workflows/library-apple.yml
@@ -78,6 +78,8 @@ jobs:
         uses: ./.github/actions/setup
         with:
           targets: aarch64-apple-ios
+      - name: Install Dependencies
+        uses: ./.github/actions/install-system-dependencies
       - name: Build Example (iOS)
         shell: bash
         run: cd apple/xcode && xcodebuild -scheme "example (iOS)" -arch arm64 -sdk iphoneos build CODE_SIGNING_ALLOWED=NO
--- a/maplibre/Cargo.toml
+++ b/maplibre/Cargo.toml
@@ -49,7 +49,7 @@ cgmath = "0.18.0"
 geo = "0.22.1"
 geo-types = { version = "0.7.6", features = ["use-rstar_0_9"] }
 rstar = "0.9.3"
-prost = "0.10.0"
+prost = "0.11.5"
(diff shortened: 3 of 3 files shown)


maxammann:
Thanks for the contribution! Yes, there is sadly an issue with the way geozero exports or doesn't export the decode functionality. 

geozero basically broke compatibility together with the release of prost. 

I'm not sure who's "fault" it is that the compilation issue was introduced.

If you want feel free to investigate why the code compiled previously but does no longer compile now.

Tristramg:
according to https://crates.io/crates/geozero/versions, the 0.9.6 was released the 2022-12-27
That release includes the bump to prost 0.11

Because `geozero = "0.9.5"` in maplibre-rs’ Cargo.toml means that 0.9.6 is ok (https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) and because there is no Cargo.lock (because it’s a binary, not a library), on my fresh install it pulled prost 0.11.0 for geozero, while a fresh install with the same code the 2022-12-26 would have pulled prost 0.10 for geozero.

I’m still a bit confused with all the traits in scope. My supposition is that because `0.10.4` and `0.11.0` are NOT compatible in a SemVer sense (while 0.9.5 and 0.9.6 are), the compiler won’t recognize the `use prost::Message;` for satisfying the required trait for `decode`.

I suppose the right fix would be that geozero re-exports its version of `prost::Message` and that you don’t need to add it as a dependency; but I can’t evaluate what that involves.

Tristramg:
Concerning the fail of the apple build: with prost 0.11, there must be a `protoc` executable available (https://github.com/tokio-rs/prost/releases/tag/v0.11.0)

Tristramg:
I made a pull request on geozero to see what their opinion is https://github.com/georust/geozero/pull/95

maxammann:
It's merged! that's great, so we can probably remove the prost dependency from our side.

Tristramg:
Indeed, once geozero makes a new release on crates.io you will be able to remove the dependency. But until then, I believe this PR is still required.

Concerning the benchmark failing, it seems that maps.tuerantuer.org is currently down, all tiles responding 404 (also on https://webgl.demo.maplibre-rs.maplibre.org/, not only the CI)

maxammann:
Thanks, yes! Let's merge this as is with admin rights! The tile server was changed and because maplibre-rs does not honor TileJSONs yet it broke.

maxammann:
Thanks for your contribution! I'll create a task for removing prost and updating geozero.

The demo is fixed again.

PR #238, opened 2023-01-14
use geo_types::Coord as geo_types::Coordinates is being deprecated

Tristramg:
see https://docs.rs/geo-types/latest/src/geo_types/geometry/coord.rs.html#33


maxammann (approved):

maxammann:
@Tristramg thanks for the contributions! :)

PR #240, opened 2023-01-15
Update trivial dependencies

Tristramg:
Edit: I had some weird troubles updating env_logger and clap on my computer. I had to `rm -r target` and rebuild from scratch. So at first I didn’t bump those, but there actually isn’t any problem.

I did not update:
* reqwest-middleware that changed its api
* rand that doesn’t seem to work with wasm target

### DIFF
--- a/android/Cargo.toml
+++ b/android/Cargo.toml
@@ -13,10 +13,10 @@ authors.workspace = true
 [dependencies]
 maplibre = { path = "../maplibre", features = ["thread-safe-futures"] }
 maplibre-winit = { path = "../maplibre-winit", version = "0.1.0"  }
-env_logger = "0.9.0"
+env_logger = "0.10.0"
 log = "0.4.17"
 ndk-glue = "0.7.0"  # version is required by winit
-jni = "0.19.0"
+jni = "0.20.0"
 
 [lib]
 #name = "maplibre_android" Currently not supported: https://github.com/rust-windowing/android-ndk-rs/issues/136
--- a/maplibre-demo/Cargo.toml
+++ b/maplibre-demo/Cargo.toml
@@ -15,10 +15,10 @@ web-webgl = ["maplibre/web-webgl"]
 trace = ["maplibre/trace"]
 
 [dependencies]
-env_logger = "0.9.0"
+env_logger = "0.10.0"
 maplibre = { path = "../maplibre", version = "0.1.0", features = ["headless", "thread-safe-futures"]  }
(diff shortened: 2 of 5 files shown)


android/Cargo.toml
```diff
@@ -13,10 +13,10 @@ authors.workspace = true
 [dependencies]
 maplibre = { path = "../maplibre", features = ["thread-safe-futures"] }
 maplibre-winit = { path = "../maplibre-winit", version = "0.1.0"  }
-env_logger = "0.9.0"
+env_logger = "0.10.0"
 log = "0.4.17"
 ndk-glue = "0.7.0"  # version is required by winit
-jni = "0.19.0"
+jni = "0.20.0"
```
maxammann, line 19: Did you test the android example? If not we might just keep the old version.

maplibre-demo/Cargo.toml
```diff
@@ -15,10 +15,10 @@ web-webgl = ["maplibre/web-webgl"]
...
 [dependencies]
-env_logger = "0.9.0"
+env_logger = "0.10.0"
 maplibre = { path = "../maplibre", version = "0.1.0", features = ["headless", "thread-safe-futures"]  }
 maplibre-winit = { path = "../maplibre-winit", version = "0.1.0"  }
 
 tile-grid = "0.3"
 
-clap = { version = "3.2.12", features = ["derive"] }
+clap = { version = "4.1.1", features = ["derive"] }
```
maxammann, line 24: I'm very much in favor of dropping clap due to frequent major version upgrades and too many features. Some simpler alternative would probably be totally enough for our demo.

Though that is probably not part of this task.

maxammann (approved):

android/Cargo.toml
Tristramg: sorry indeed, I went a bit too far.
I haven’t yet setup the android toolchain. I switched it bach to 0.19

maplibre-demo/Cargo.toml
Tristramg: structopt used to be my goto library, but it was based on clap, and got finally completely merged into clap.

It’s indeed a very big library, but when using simpler ones, there are rapidly small annoyances that aren’t managed.

There are options listed here https://github.com/rosetta-rs/argparse-rosetta-rs

maxammann (approved):

maplibre-demo/Cargo.toml
maxammann: I'll maybe test out argh one day.

clap is just so big for simple tasks :)

PR #241, opened 2023-01-15
Simplify and move Stencil reference functions

Tristramg:
This pull request is more a question of style concerning `stencil_reference_value_2d` and `stencil_reference_value_3d`

The first commits changes a bit how the pattern matching is written. I believe it’s a bit more “rusty” and we have the guarantee the matching is complete.

The second commit moves those functions as a member of `WorldTileCoord` instead of `TileViewPattern`. I’m still missing the bigger picture of all those parts. The observation was that no information of `TileViewPattern` is required to do the computation, but I’ll perfectly understand that there are broader implications.

### DIFF
--- a/maplibre/src/coords.rs
+++ b/maplibre/src/coords.rs
@@ -457,17 +457,6 @@ impl WorldTileCoords {
         })
     }
 
-    /// 2D version of [`TileViewPattern::stencil_reference_value_3d`]. This is kept for reference.
-    /// For the 2D case we do not take into account the Z value, so only 4 cases exist.
-    pub fn stencil_reference_value_2d(&self) -> u8 {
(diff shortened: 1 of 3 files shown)


maxammann:
Historically they moved a little bit around in code. So thanks for moving them to the right location.

I think the 2d is not used anymore. I think this would be a good PR to just remove it and document via a commit why it was removed. Here is why:

`The 2d version of calculating the stencil value is only valid for cases where only tiles from exactly one zoom level are displayed at once. As soon as multiple zoom levels are presented at once, the stencil value needs to be distinct across zoom levels`.

Can you remove it and commit with that message?

maxammann (dismissed):

Tristramg:
Function removed! Thank you for the context

PR #249, opened 2023-01-29
Use prost Message from geozero

maxammann:
@Tristramg this fixes the issue you discovered.


Tristramg:
Nice to see that there was a quick relesase!

Issue #92, opened 2022-05-09
Remove all `.unwrap()` and `.expect()`


maxammann:
## 🤔 Expected Behavior

Return a Result for failable operations.

## 😯 Current Behavior
We use sometimes unwrap or expecte because of lazyness.

maxammann:
Hello @iakev I'm merging in a rather large PR in the next 1h :) So its better to start working after that.
Apart from that you should be fine.

Maybe it could make sense to leave some `.expect()` if they are very unlikely, or if you are unlikely to hit them after setting up the map.

maxammann:
Or probably tomorrow..

iakev:
Okay, no problem will start working on it tomorrow. In case I find any
difficulties, I presume I can contact you right?

maxammann:
Yes, absolutely. You can also reach me on matrix.org Its @maxammann

iakev:
Hello, I hope you are fine. So I have tweaked one file just wanted to hear
if what am doing is correct so that I can continue. Can I push the branch
for your review?

Best Regards,
Kevin Mwongera

maxammann:
Sure, just push the branch to your fork and then open a PR in "Draft Mode": https://github.blog/2019-02-14-introducing-draft-pull-requests/

iakev:
Thanks, I will do that. Kindly let me know your thoughts.

xinslu:
Hi,
Is this issue still open? If yes, then I'd love to get involved!

maxammann:
Hi! The description and goal of this issue needs some refinement. Ill try to refine it this weekend!

PR #261, opened 2023-02-19
Remove unwrap from write_png

Tristramg:
Unwraps are bad, and we should gradually remove them (#92) 

This pull request removes them from a single function to see if we agree on how to do it.

The function now returns a Result (two specific errors have been created) and there still is an `expect`. So the in case of a failure, the program will still panic.

## 🚨 Test instructions

`cargo run -p maplibre-demo  --features headless -- headless`

It write 4 png on the current path

### DIFF
--- a/maplibre/src/headless/system.rs
+++ b/maplibre/src/headless/system.rs
@@ -55,10 +55,12 @@ impl System for WriteSurfaceBufferSystem {
                 let padded_buffer = buffer_slice.get_mapped_range();
 
                 if self.write_to_disk {
-                    buffered_texture.write_png(
-                        &padded_buffer,
-                        format!("frame_{}.png", current_frame).as_str(),
-                    );
+                    buffered_texture
+                        .write_png(
+                            &padded_buffer,
+                            format!("frame_{}.png", current_frame).as_str(),
+                        )
+                        .expect("Could save frame to disk");
                 }
 
                 // With the current interface, we have to make sure all mapped views are
--- a/maplibre/src/render/resource/surface.rs
+++ b/maplibre/src/render/resource/surface.rs
@@ -1,8 +1,9 @@
 //! Utilities for handling surfaces which can be either headless or headed. A headed surface has
 //! a handle to a window. A headless surface renders to a texture.
 
-use log::debug;
 use std::{mem::size_of, num::NonZeroU32, sync::Arc};
+
+use log::debug;
 #[cfg(feature = "headless")]
 use thiserror::Error;
 
--- a/maplibre/src/render/resource/surface.rs
+++ b/maplibre/src/render/resource/surface.rs
@@ -1,9 +1,10 @@
 //! Utilities for handling surfaces which can be either headless or headed. A headed surface has
 //! a handle to a window. A headless surface renders to a texture.
 
-use std::{mem::size_of, num::NonZeroU32, sync::Arc};
-
 use log::debug;
+use std::{mem::size_of, num::NonZeroU32, sync::Arc};
+#[cfg(feature = "headless")]
+use thiserror::Error;
 
 use crate::{
     render::{eventually::HasChanged, resource::texture::TextureView, settings::RendererSettings},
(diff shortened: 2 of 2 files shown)


maplibre/src/headless/system.rs
```diff
@@ -55,10 +55,12 @@ impl System for WriteSurfaceBufferSystem {
...
-                    buffered_texture.write_png(
-                        &padded_buffer,
-                        format!("frame_{}.png", current_frame).as_str(),
-                    );
+                    buffered_texture
+                        .write_png(
+                            &padded_buffer,
+                            format!("frame_{}.png", current_frame).as_str(),
+                        )
+                        .expect("Could save frame to disk");
```
maxammann, line 63: I see two possible solutions here:

- Accept that some excepts are fine. For instance wgpu is also panicing for some inputs.
- Systems should be able to return errors, alternatively, the errors need to be sent somewhere.

maplibre/src/render/error.rs
```diff
@@ -10,6 +10,11 @@ pub enum RenderError {
     Graph(#[from] RenderGraphError),
     #[error("error while requesting device")]
     RequestDevice(#[from] wgpu::RequestDeviceError),
+    #[cfg(feature = "headless")]
+    #[error("error while writing surface to png")]
+    WritePng(#[from] png::EncodingError),
+    #[error("error in file io")]
+    IoError(#[from] std::io::Error),
```
maxammann, line 17: Instead of adding these errors to the RenderError, a new error type should be created: HeadlessError/HeadlessRender error.

That also eliminates the need for the `cfg()`

maxammann: Why this is good idea is described here: https://mmapped.blog/posts/12-rust-error-handling.html#errors-problem-vs-solution

maplibre/src/render/resource/surface.rs
```diff
@@ -98,32 +98,32 @@ impl BufferedTextureHead {
         self.output_buffer.unmap();
     }
 
-    pub fn write_png<'a>(&self, padded_buffer: &wgpu::BufferView<'a>, png_output_path: &str) {
+    pub fn write_png<'a>(
+        &self,
+        padded_buffer: &wgpu::BufferView<'a>,
+        png_output_path: &str,
+    ) -> Result<(), crate::render::error::RenderError> {
```
maxammann, line 105: Ideally errors are kept very local. So at least for every `impl {}` block/struct, and maybe even for every function a new error type should be created.

I'm trying to follow the guidelines here :) https://mmapped.blog/posts/12-rust-error-handling.html#errors-problem-vs-solution

maxammann (changes requested):

maxammann:
> Unwraps are bad, and we should gradually remove them (#92)

Absolutely! This is usually the thing I ignore while doing major refactors because it is trival to do later on in separate clean PRs :)

I'm welcoming the effort!

Ideally we get rid of all .unwrap() and most .expect().

maplibre/src/headless/system.rs
Tristramg: I would say that on a longer run, it will be nicer to move the errors up and that system return errors.

But the current state seems acceptable to me: if a headless render can’t save to disk, it’s ok to fail. (but maybe I’m too much baby steps oriented?)

maplibre/src/render/error.rs
Tristramg: I moved it separately. May you should have a look at the naming. I tried to make something more specific that `IoError`

Tristramg:
Getting rid of expects is for me a nice way to get a bit deeper, and getting feedback on how things are (or should be) organized and can be done with incremental steps

maxammann:
I'd actually keep the error even more local: https://github.com/maxammann/maplibre-rs/commit/1a80067ffef104240305a5872cbcbba100126c7a

I'm happy to discuss this because this is a general question which should be answered for this project :)

Tristramg:
It’s ok for me. I might need some time to overcome my tendency to define all the errors in one specific place. I am convinced by the argumentation, but it’s not yet in my muscle memory ;)

I’ve rebased it with you suggestion

maplibre/src/render/resource/surface.rs
```diff
@@ -1,9 +1,10 @@
 //! Utilities for handling surfaces which can be either headless or headed. A headed surface has
 //! a handle to a window. A headless surface renders to a texture.
 
-use std::{mem::size_of, num::NonZeroU32, sync::Arc};
-
 use log::debug;
+use std::{mem::size_of, num::NonZeroU32, sync::Arc};
+#[cfg(feature = "headless")]
```
maxammann, line 6: minor remark: I usually just write out the paths where it is used instead of using/import.

maxammann (approved):
Looks good, merging!

maxammann:
I just formatted the code again :)

maplibre/src/render/resource/surface.rs
Tristramg: :+1:  thank you for pushing the fix ;)

PR #265, opened 2023-02-22
Styleparser

kimpham54:
Working on the style parser that can parse jsons based on the maplibre style specification https://maplibre.org/maplibre-gl-js-docs/style-spec/ and https://github.com/maplibre/maplibre-gl-js/blob/main/src/style-spec/reference/v8.json. 

- Added properties from the style spec
- Added 'extra' capturing properties that weren't captured in the style spec
- Added tests to run against actual style jsons
- You don't have to include the work from the map-styles, original, and tests directory

## 🚨 Test instructions

`cargo test --package maplibre --lib -- style::style::tests` shows that 30 tests are passing, 2 are failing. I need to work on getting those two tests to pass - currently the two styles that fail have an array with a mix of integers and floats, they need to be parsed in a way that can accept both types


maxammann:
@kimpham54 I removed the proprietary files and removed the tests. We can create those tests in some other repository hosted outside of maplibre.

For that we need to expose the style parsing functionality. Either through a new maplibre-style crate or by exporting the functions from the existing maplibre crate.

Reviewing now :)

maplibre/src/style/expression.rs
maxammann, line 1: Remove this file

maplibre/src/style/.gitignore
maxammann, line 1: Remove this file

maplibre/src/style/original/layer.rs
```diff
@@ -0,0 +1,97 @@
...
+use cint::{Alpha, EncodedSrgb};
+use csscolorparser::Color;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct BackgroundPaint {
+    #[serde(rename = "background-color")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub background_color: Option<Color>,
+    // TODO a lot
```
maxammann, line 14: These are still do do. Have you investigated how much work this is?

maplibre/src/style/source.rs
```diff
@@ -45,8 +45,14 @@ pub struct VectorSource {
     /// Array of URLs which can contain place holders like {x}, {y}, {z}.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub tiles: Option<TileUrl>,
-    // url: Option<TileJSONUrl>,
+
+    // #[serde(skip_serializing_if = "Option::is_none", rename = "testtype")]
+    // pub type: Option<String>,
```
maxammann, line 50: What about these lines?

maplibre/src/style/source.rs
```diff
@@ -56,4 +62,6 @@ pub enum Source {
     Vector(VectorSource),
     #[serde(rename = "raster")]
     Raster(VectorSource), // FIXME: Does it make sense that a raster have a VectorSource?
```
maxammann, line 64: Indeed, we should rename VectorSource.

maplibre/src/style/style.rs
```diff
@@ -16,117 +20,243 @@ use crate::style::{
...
+    pub layers: Vec<StyleLayer>,
+
+    // TODO this is metadata, and it's not in the spec
+    pub created: Option<String>,
+    pub modified: Option<String>,
+    pub id: Option<String>,
+    pub owner: Option<String>,
+    pub visibility: Option<String>,
+    pub protected: Option<bool>,
+    pub draft: Option<bool>,
```
maxammann, line 48: If this not in the spec, we might put this data into a Map<String, String>

maxammann: But maybe its better to have these here, in case they exist we can just parse them.

maplibre/src/style/style.rs
```diff
@@ -16,117 +20,243 @@ use crate::style::{
...
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Default)]
+pub struct Metadata {
+    Value: String,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Default)]
+pub struct Light {
+    #[serde(default)]
```
maxammann, line 62: 👍🏼

maplibre/src/style/style.rs
```diff
@@ -16,117 +20,243 @@ use crate::style::{
...
+    fn default() -> Self {
+        Color("#ffffff".parse().unwrap())
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+struct Intensity(Vec<f32>);
+impl Default for Intensity {
+    fn default() -> Self {
+        Intensity(vec![1.15, 210.0, 30.0])
```
maxammann, line 84: It would be super awesome to link to the maplibre style spec e.g. https://maplibre.org/maplibre-style/layers/#minzoom

This is actually true for all properties. They should preferably link to the docs.

maplibre/src/style/style.rs
```diff
@@ -16,117 +20,243 @@ use crate::style::{
...
+            extra: Default::default(),
             version: 8,
             name: "Default Style".to_string(),
             metadata: Default::default(),
             sources: Default::default(),
             center: Some([46.5197, 6.6323]),
             pitch: Some(0.0),
             zoom: Some(13.0),
+            bearing: Some(0.0),
+            created: Some("2021-03-01T00:00:00Z".to_string()),
```
maxammann, line 146: Better put here something like 1970-01-01

maplibre/src/style/style.rs
```diff
@@ -1,10 +1,14 @@
 //! Default vector tile styles configuration.
+// use std::env;
 
-use std::{collections::HashMap, str::FromStr};
+use std::{collections::HashMap, io, path::Path, str::FromStr};
 
-use csscolorparser::Color;
+use csscolorparser::Color as CssColor;
```
maxammann, line 6: It would be good if we could limit the dependencies. Is it hard to implement this crate on our own. What does the maplibre spec say about color parsing?

maplibre/src/style/tests/test-data/cl4.json
maxammann, line 1: Did you create this file or did you copy it from somewhere?

maplibre/src/style/tests/test-data/sample1.json
maxammann, line 1: Did you create this file or did you copy it from somewhere?

maplibre/src/style/tests/test-data/sourcedraft.rs
maxammann, line 1: Is this file used? Why is it in test-data?

maxammann: Not reviewed.

maplibre/src/style/style.rs
```diff
@@ -140,12 +270,30 @@ impl Default for Style {
...
             ],
         }
     }
 }
 
+#[derive(Debug)]
+pub enum StyleModuleError {
+    Io(io::Error),
+    Json(serde_json::Error),
+}
```
maxammann, line 284: Use here https://crates.io/crates/thiserror

We might not want to pass through serde_json here so we can later more easily replace serde with something else.

maplibre/src/style/original/source.rs
maxammann, line 1: I think these files in orignal can be removed.

maxammann (changes requested):
Overall the direction is good. I like the Default impls which follow the spec.

I think the best way moving forward is to create a maplibre-style crate in the root dependency. Then you copy the source.rs and style.rs files (and others) to that new create.

You can then revert he maplibre/src/style directory to the state it was in before this PR. When the maplibre-rs create is more finished, then we can see how we integrate it into the maplibre crate again.

nyurik:
Lots of amazing work on this PR, thanks @kimpham54 !  I wonder if we can point this work to https://github.com/maplibre/maplibre-gl-js/tree/main/test/integration/render/tests -- it has a lot of MapLibre style files that could be used for testing?  Also, any let me know if I can help moving this along.

HarelM:
I think this is a great step forward!
The way I imagined it was that all the parsing and validation of the style spec would be in the maplibre-style-spec repo, creating a packages for every platform.
This way all the style.json files that needs to be tested are there as well.
Later on, if we decide for example to change the underlying architecture to use json scheme it will all be in the same place.
This is just an initial thought though...

maxammann:
@kimpham54 Hi! are you still interested in pushing this forward? :)

PR #273, opened 2023-04-11
Increase coords consistency and clamp zoom level to numerically feasible range

terrorfisch:
- Clamp the integer `ZoomLevel` to `0..31` when converted from floating point `Zoom`
 - Restrict traversal of `WorldTileCoords` to supported zoom level range
 - Construct QuadKey with the [`morton_encoding`](https://docs.rs/morton-encoding/latest/morton_encoding/) crate.

Fixes crash from #161 by making the "backend" more robust.

Does *not* implemented the desired zoom limit. This is done in #180 

## 💻 Examples

You can arbitrarily zoom in/out in the demo without panic.

## 🚨 Test instructions

Launch demo and start zooming :)

## ✔️ PR Todo

I am not satisfied with these points yet and lack the domain knowledge to proceed:

- [ ] Check `get_children` usage in renderer
- [ ] Properly handle signedness of `WorldTileCoords`
- [ ] Check if saturation in `ViewRegion::iter` is the correct way to go


maplibre/src/coords.rs
```diff
@@ -22,54 +22,50 @@ pub const EXTENT_UINT: u32 = 4096;
...
 
 impl Quadkey {
-    pub fn new(quad_encoded: &[ZoomLevel]) -> Self {
-        let mut key = [ZoomLevel::default(); MAX_ZOOM];
-        key[0] = (quad_encoded.len() as u8).into();
-        for (i, part) in quad_encoded.iter().enumerate() {
-            key[i + 1] = *part;
+    pub fn new(zoom_level: ZoomLevel, x: u32, y: u32) -> Self {
+        // use morton enconding / Z-ordering to build linear quadtree index
+        let encoded_coords = morton_encoding::morton_encode([x, y]);
```
maxammann, line 42: Interesting approach. As this is used a lot, can we benchmark the encoding/decoding OR do you have a guess on how fast this is?

maplibre/src/coords.rs
```diff
@@ -22,54 +22,50 @@ pub const EXTENT_UINT: u32 = 4096;
...
-        }
-        Ok(())
+        let zoom_level = self.zoom_level;
+        let [x, y]: [u32; 2] = morton_encoding::morton_decode(self.encoded_coords);
+
+        f.debug_struct("QuadKey")
+            .field("zoom_level", &zoom_level)
+            .field("x", &x)
+            .field("y", &y)
+            .finish()
```
maxammann, line 60: Cool! Actually I was not aware of this API!

maplibre/src/coords.rs
```diff
@@ -82,36 +78,45 @@ impl fmt::Debug for Quadkey {
...
+            None
+        } else {
+            Some(Self(self.0 - 1))
+        }
+    }
 
-    fn sub(self, rhs: u8) -> Self::Output {
-        let zoom_level = self.0.checked_sub(rhs).expect("zoom level underflowed");
-        ZoomLevel(zoom_level)
+    pub(crate) fn as_u8(&self) -> u8 {
```
maxammann, line 118:
```suggestion
    pub fn as_u8(&self) -> u8 {
```

So far I'm not using pub(crate) anywhere.

maxammann: Should we implement Deref instead of as_u8?

maxammann (commented):
Still have to go through all, but looks cool!

Did you verify by running this?

maplibre/src/coords.rs
terrorfisch: It compiles to ~20 branchless instructions of x86 and should be really fast: https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=6693a9f5ac38027b1fefe470f4208428 (select ASM in the ... menu on the top left)

maplibre/src/coords.rs
terrorfisch: Imo `Deref`` counteracts the new type idea. I like the explicitness more.

I used `pub(crate)` because this feels very "internal" to me but I guess the question what will be stable API and what wont be is not really relevent yet.

terrorfisch:
> Did you verify by running this?

Yes. Works on my Windows 10 machine.

maplibre/src/coords.rs
```diff
@@ -22,54 +22,50 @@ pub const EXTENT_UINT: u32 = 4096;
...
-pub const MAX_ZOOM: usize = 32;
 
-// FIXME: MAX_ZOOM is 32, which means max bound is 2^32, which wouldn't fit in u32 or i32
-// Bounds are generated 0..=31
-pub const ZOOM_BOUNDS: [u32; MAX_ZOOM] = create_zoom_bounds::<MAX_ZOOM>();
+/// Maximal zoom level. Typically, the maximal zoom is 18 as described
+/// [here](https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames#Zoom_levels).
+/// This implementation allows 30 because this covers all tile sizes that can be represented
+/// with signed 32-bit integers.
+pub(crate) const MAX_ZOOM: u8 = 30;
```
terrorfisch, line 30: This should probably be an associated constant of `ZoomLevel`, i.e. `ZoomLevel::MAX`.

maplibre/src/coords.rs
```diff
@@ -82,36 +78,45 @@ impl fmt::Debug for Quadkey {
...
+        }
     }
     pub fn is_root(self) -> bool {
         self.0 == 0
     }
-}
 
-impl std::ops::Add<u8> for ZoomLevel {
-    type Output = ZoomLevel;
+    pub(crate) fn max_tile_coord(&self) -> u32 {
```
maxammann, line 98:
```suggestion
    pub fn max_tile_coord(&self) -> u32 {
```

maplibre/src/coords.rs
```diff
@@ -390,70 +392,49 @@ impl WorldTileCoords {
...
-
-        if x >= bounds || y >= bounds {
-            return None;
-        }
-
-        let mut key = [ZoomLevel::default(); MAX_ZOOM];
+        // check for out of bound access
+        let TileCoords {
+            x, y, z
+        } = self.into_tile(TileAddressingScheme::XYZ)?;
```
maxammann, line 399: Quadkeys should work on WorldTileCoords. Hardcoding here the addressing scheme seems odd

maplibre/src/render/tile_view_pattern/mod.rs
```diff
@@ -136,7 +136,7 @@ pub trait HasTile {
                 if self.has_tile(child, world) {
                     output.push(child);
                 } else {
-                    new_children.extend(child.get_children())
+                    new_children.extend(child.get_children().into_iter().flatten())
```
maxammann, line 139:
```suggestion
                    new_children.extend(child.get_children()?)
```

maxammann (approved):
Looks good overall! I guess there is no important rationale in these changes. If there is one document WHY those changes are necessary.

Happy to accept this though, as it probably helps you to get into the project!

maxammann:
Format can be fixed with `just fmt`

maplibre/src/render/tile_view_pattern/mod.rs
terrorfisch: This would be an early return if one child has no children. Is this desired?

maplibre/src/coords.rs
terrorfisch: The only purpose here is to reuse the bound check code. Quite hacky. I will refactor it a bit to make that clear.

maplibre/src/render/tile_view_pattern/mod.rs
maxammann: Ah got it, sorry! :)

nyurik:
Rebased on main, and cargo fmt-ed it

maxammann:
@terrorfisch I updated this to the lastest main. If you want to go over it again, then I can review this one.

Regardin the morton-encoding dependency: I think it would be nice to include the efficiency aspect (compiled down to 20 instructions) in the source code as comment.

maxammann:
@terrorfisch Same here, I would still love to merge this :)

PR #279, opened 2023-05-24
Fix Apache license file

nyurik:
This makes github detect license correctly.
Note that our Martin, as well as Rust lang itself has exactly this approach

| this is how github main pages looks at the moment (right side) |
|---|
| image |

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -1,13 +1,176 @@
-Copyright 2022 MapLibre contributors
+                              Apache License
+                        Version 2.0, January 2004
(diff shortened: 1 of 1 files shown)


maxammann (approved):

LICENSE-APACHE
```diff
@@ -1,13 +1,176 @@
-Copyright 2022 MapLibre contributors
```
birkskyum, line 1: @nyurik , is it possible to add back the "Copyright 2022 MapLibre contributors" and have github recognize apache like to does MIT?

PR #280, opened 2023-05-24
Format justfile using just's own --fmt

nyurik:
Run `just --fmt --unstable`

When reviewing, I suggest viewing it [without whitespace changes](https://github.com/maplibre/maplibre-rs/pull/280/files?diff=split&w=1)

### DIFF
--- a/justfile
+++ b/justfile
@@ -4,99 +4,96 @@
 
 set shell := ["bash", "-c"]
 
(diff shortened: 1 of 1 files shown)


maxammann (approved):

maxammann:
Thanks I was not aware of it!

maxammann:
Hm, we would add just --fmt to the Justfile's fmt script

PR #281, opened 2023-05-28
chore: Consolidate dependency version mgmt

nyurik:
Move all dependencies to workspace so that we can keep all versions across packages in one spot.

Also add a geo keyword to the packagess.

See https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table


maxammann:
Merging with a 99% chance in 10h :) Let's sleep over this :D

maxammann (approved):

PR #286, opened 2023-09-22
Chore: upgrade deps

nyurik:
* rename `keywords` into `categories` (because that's what it was used for), and added a new `keywords`
* bump nightly and stable rust versions
* run `just fmt`
* remove backends config for wgpu (dep no longer uses that param)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,9 +19,11 @@ members = [
 description = "Maps for Desktop, Mobile and Web"
 version = "0.1.0"
 edition = "2021"
-rust-version = "1.65"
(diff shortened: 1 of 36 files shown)


birkskyum (changes requested):
Tests appear to be failing.

maxammann:
Thanks! I was about to do the same as Eurorust prep.

I think rust must be updated

nyurik:
@maxammann and/or @birkskyum  plz take a look

maxammann:
Finally formatting for let else :)

maxammann (approved):

Issue #298, opened 2024-03-08
Demo will not compile, issue with cargo dependencies: clap v4.5.1


96fps:
Cloned the repo, ran `cargo run -p maplibre-demo`
```
error: package `clap v4.5.1` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.73.0
Either upgrade to rustc 1.74 or newer, or use
cargo update -p clap@4.5.1 --precise ver
where `ver` is the latest version of `clap` supporting rustc 1.73.0
```

This copied from a remote Ubuntu machine, similar error in WSL2 and via Windows 11 (with git bash)

It's not platform specific but due to rust package management. The `Cargo.toml` seems to specify version v4.3.1 of this package, so why is cargo fetching an incompatible version 4.5.1?

96fps:
Updating the version of rust on `rst-toolchain.toml` from 1.7.3 to 1.7.6 seems to resolve this.

maxammann:
Oh oke annoying. I think it would make sense to commit the Cargo.toml for the time being.

Actually this is a bug in clap because a minor version update should not break things like that imo.

nyurik:
And that's why I find [cargo semver checker](https://github.com/obi1kenobi/cargo-semver-checks?tab=readme-ov-file#quick-start) so important as part of the CI.  Two lines in CI - save tons of pain for everyone.

PR #299, opened 2024-03-09
Update Cargo.toml to fix dependency conflict

96fps:
Addresses issue: [289](https://github.com/maplibre/maplibre-rs/issues/298)

"clap" v4.5.2 does not compile with the defined toolchain and rust version. specifying exact version of dependency instead.

Tested by compiling the demo.


maxammann:
Thanks for the fix, let's just upgrade Rust however :) https://github.com/maplibre/maplibre-rs/pull/300

PR #301, opened 2024-03-13
Fix builds

maxammann:
This should fix:
* NDK version
* version of wasm-bindgen
* fix the nix-shell for usage on Nix


nyurik (approved):
lgtm

PR #311, opened 2024-07-23
Upgrade wgpu to 22

maxammann:


maplibre/src/render/mod.rs
```diff
@@ -492,6 +501,7 @@ mod tests {
                     label: None,
                     required_features: wgpu::Features::default(),
                     required_limits: wgpu::Limits::default(),
+                    memory_hints: wgpu::MemoryHints::default(),
```
nyurik, line 504: why not `&wgpu::DeviceDescriptor::default()` ?

nyurik (approved):
minor optional nit

maplibre/src/render/mod.rs
maxammann: I think this minor dirtyness is kind of helpful for me, so I notice when new props are added and I can see more clearly what can be configured. 

But a lint for this would be cool, right?

nyurik: hmm, good point, i wonder if clippy could catch things like that...  but then it may start suggesting things like `replace 0 with i32::default()` :D

PR #318, opened 2024-08-29
Update ndk version

maxammann:
GH Actions removed older NDKs


maxammann:
see https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md

PR #335, opened 2026-04-02
fix: typos

mwtoews:
Fix a few typos identified with codespell.

Unsure if changes to `maplibre/src/legacy/*` is welcome; these can be rolled-back if to be kept truly legacy.

### DIFF
--- a/maplibre/src/legacy/shaping.rs
+++ b/maplibre/src/legacy/shaping.rs
@@ -339,14 +339,14 @@ fn get_glyph_advance(
             return 0.0;
         }
 
-        if it.expect("cant be none").is_none() {
+        if it.expect("can't be none").is_none() {
             return 0.0;
         }
 
         return (it
-            .expect("cant be none")
+            .expect("can't be none")
             .as_ref()
-            .expect("cant be none")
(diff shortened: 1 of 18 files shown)


CommanderStorm (approved):
Thank you

CommanderStorm (dismissed):

maplibre/src/legacy/shaping.rs
```diff
@@ -705,20 +705,20 @@ fn shape_lines(
...
                     }
 
-                    if glyph.expect("cant be none").is_none() {
+                    if glyph.expect("can't be none").is_none() {
                         continue;
                     }
 
-                    metrics =
-                        (glyph.expect("cant be none").as_ref().expect("cant be none")).metrics;
+                    let glyphs = glyph.expect("can't be none").as_ref().expect("can't be none");
```
CommanderStorm:
```suggestion
                    let glyphs = glyph
                        .expect("can't be none")
                        .as_ref()
                        .expect("can't be none");
                    metrics = glyphs.metrics;
```

CommanderStorm (approved):

Issue #76, opened 2022-05-04
Support S3/https/file protocols


maxammann:
We would like to be able to use different protocols for retching mbtiles/pmtiles/tiles.

## 🤔 Expected Behavior

When support in PM tiles the scenario would look like that:

![image](https://user-images.githubusercontent.com/905221/166680575-3016908d-cb7c-4df3-99e8-dd54acdf3d59.png)

We would like to support multiple protocols between maplibre and hosted data:

![image](https://user-images.githubusercontent.com/905221/166680687-0a522492-1f26-4ed6-afbb-e7e87d567c2c.png)

## 😯 Current Behavior

Current scenario:
![image](https://user-images.githubusercontent.com/905221/166680508-65b3eac8-15de-4bb1-a200-435322f54d17.png)

PR #338, opened 2026-08-10
Add offline MBTiles source for Apple

luofang34:
## Summary

- add a read-only MBTiles source client for native targets
- inject the same local source into main and worker kernels
- expose the source through the Apple C facade and Swift wrapper
- keep SQLite inside the static Apple framework
- preserve JSON style layer order

The client converts XYZ rows to TMS rows and decompresses gzip tile payloads. This path does not create an HTTP client.

This change implements the local file part of #76.

## Validation

- just fmt-check
- cargo test -p maplibre --lib: 39 passed, 2 ignored
- cargo test -p apple --lib: 4 passed
- cargo clippy --no-deps -p apple --target aarch64-apple-ios
- cargo clippy --no-deps -p maplibre-winit
- unsigned Xcode iOS example build with the iOS 27 SDK
- macOS probe read a reviewed MBTiles archive and emitted vector draw commands

## iOS runtime limit

The current winit Apple path uses the legacy application lifecycle. The iOS 27 runtime stops that path before renderer startup. This PR does not change the application lifecycle. The local source adapter and Apple build are independent of that upstream lifecycle work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ android_logger = "0.14.1"
 png = { version = "0.17.10" }
 reqwest = { version = "0.12.5", default-features = false, features = ["rustls-tls", "gzip"] }  # Use rusttls on android because cross compiling is difficult
 rstar = "0.12.0"
-rusqlite = { version = "0.32.0" }
+rusqlite = { version = "0.32.0", features = ["bundled"] }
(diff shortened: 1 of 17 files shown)


luofang34:
Runaway AI slop. Apologize and please ignore🥲

